PR #2, opened 2019-01-08
Make output format field non case-sensitive

firecrafty:


mjansen4857:
This was an issue I totally missed because of my personal settings. I already made commits to fix this and released a patch.

PR #3, opened 2019-01-18
Add x and y positions as output format options

bob80333:
Some path following algorithms need the x and y coordinates


mjansen4857:
Good idea. One thing to note is that the x and y coordinates are not field oriented. The origin seems to be the top center of the robot.

PR #23, opened 2020-01-08
Constrain the main point of a spline on the Y-axis if the shift button is held down

DeanDavis:
Added code to constrain the main point of a spline on the Y-axis if the shift button is held down.
This is similar to constraining the angle of the offset control points of a spline.

Also, my code editor made some cosmetic code formatting changes as well. Sorry.


mjansen4857:
Seems useful. I might mess with it at some point to make it work for the X-axis as well but you're welcome to try too if you get to it before I can.

PR #72, opened 2022-03-18
Support adding a waypoint in the middle of a path

### LINKED ISSUE #56: Add a Waypoint in the middle of a Path
It currently supports only adding a waypoint at the end of the path.
Would be helpful to add waypoints in the middle of the path.

realSaddy:
Good day,

This PR changes the generation of waypoints. New waypoints are added via a double click from the last selected waypoint. This allows new points to be added mid-path, which is very useful for adding changes to the start of a long path.

Best,
Jack
Programming Lead - Texas Torque (FRC 1477)

Closes #56

### DIFF
--- a/lib/robot_path/robot_path.dart
+++ b/lib/robot_path/robot_path.dart
@@ -68,15 +68,29 @@ class RobotPath {
     print('Saved and generated path in ${s.elapsedMilliseconds}ms');
   }
 
-  void addWaypoint(Point anchorPos) {
-    waypoints[waypoints.length - 1].addNextControl();
-    waypoints.add(
-      Waypoint(
-        prevControl:
-            (waypoints[waypoints.length - 1].nextControl! + anchorPos) * 0.5,
+  void addWaypoint(Point anchorPos, int waypoint) {
+    if(waypoints[waypoint].nextControl == null) { 
+      print("Adding waypoint at the end");
+      waypoints[waypoints.length-1].addNextControl();
+      waypoints.add(
+        Waypoint(
+          prevControl:
+            (waypoints[waypoints.length-1].nextControl! + anchorPos) * 0.5,
         anchorPoint: anchorPos,
-      ),
-    );
+        ),
+      ); 
+    } else {
+      print("Adding waypoint in the middle of the path");
+      waypoints.insert(
+        waypoint+1, Waypoint(
+          prevControl:
+            (waypoints[waypoint].anchorPoint + anchorPos) * 0.5,
+          nextControl:
+            (waypoints[waypoint].nextControl! + anchorPos) * 0.5,
+          anchorPoint: anchorPos,
+        )
+      );
+    }
   }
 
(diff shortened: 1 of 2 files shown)


mjansen4857:
This sounds a bit more elegant that the way I planned on doing it. Thanks. Does it still default to adding a point at the end of the path if no waypoint is selected?

realSaddy:
It does not at the moment. Let me add that in.

realSaddy:
@mjansen4857 Added.

mjansen4857:
It seems to not be setting both control points correctly when adding a point in the middle:
image

Easiest way to fix this is just to get the difference between the anchor point and the previous control point and then add that difference to the anchor point to get the position of the next control point. For example:
```
prevControl = (1, 1)
anchorPoint = (3, 2)

deltaX = 3 - 1 = 2
deltaY = 2 - 1 = 1

nextControl = (3 + 2, 3 + 1) = (5, 4)
```

realSaddy:
Good catch. We use holonomic so I was only testing it through that. Will fix this and notify you when done.

realSaddy:
@mjansen4857 Sorry about all the bugs. Please let me know if the latest commit works for you.

mjansen4857:
No problem. Bugs are part of the process. I found a couple more issues. When you first open the app and immediately try to add a new waypoint it will add it after the start point instead of the end. It should add it to the end here since no waypoint is selected. I'm assuming the selected index variable is not being initialized correctly. The second issue is that when no waypoint is selected, double clicking does not add a new waypoint most of the time. It seems to be random when it will actually add a waypoint after a double click, but after it does it the first time it works every time after that until I switch to a different path or something. I have absolutely no idea what could be causing that. If I switch back to the main branch this doesn't happen anymore so the issue is somewhere in the code for this PR.

realSaddy:
@mjansen4857 I have fixed the issue with it not adding to the end after startup.

I cannot replicate the issue you describe. Is there any console output that could help? I am using Linux; perhaps it is a Windows-only issue?

mjansen4857:
It is printing this error: `Another exception was thrown: RangeError (index): Invalid value: Not in
inclusive range 0..2: 3`
The stack trace leads to robot_path.dart line 72. So it is trying to access the waypoint with index 3 in a list of length 3 so its out of bounds. I think I actually tracked down the cause. It seems to only happen after an undo. I think you can reproduce it if you add a point to the end of a path, undo, and then try to add it again. I believe the selected index is not being changed after an undo is performed. After an undo, no waypoint is selected so the selected index should reflect this as well.

mjansen4857:
That seems to have fixed it. Just one more little nitpick thing: I think it would be better and avoid other possible issues if you changed any line setting the selected index to waypoints.length-1 to just setting the selected index to -1. This would prevent having a specific index stored unless one is actually selected. You already have a check to consider the last waypoint as selected whenever the selected index is -1 so its better to just rely on that check.

realSaddy:
@mjansen4857 :+1:

PR #80, opened 2022-03-29
Add ability to do math in number text boxes

carelesshippo:
This adds the ability to evaluate math equations in the text box so you can now do something like `180 / 2` or `213 - 90` and such. It uses the function tree library. This is especially useful for me with transformations when creating paths and now you don't have to keep a calculator open. I'm not very familiar with Dart so I very well may have done something wrong and the regex could very well not be the greatest.


mjansen4857:
Looks good. Just have a couple suggestions to make things a little more concise:
* The regex can be made a little shorter by using: `(^(-?)\d*\.?\d*)([+/\*\-](-?)\d*\.?\d*)*`. They do the exact same thing so it doesn't really matter which one is used.

* Line 264 of waypoint_card.dart can be changed to `num parsed = val.interpret();`

PR #99, opened 2022-08-02
Fix velocity calculation in C++ version.

bckircher:
When I changed the calculation of curvature to make it be both positive and negative based on the direction the robot is turning, I missed an absolute value on the computation of the maximum curvature velocity in the C++ version (though I got it correct on the Java version, and it got translated correctly into the Dart version). This adds the missing absolute value so that the velocity correctly slows down around sharper turns.


mjansen4857:
I thought I remembered seeing this before and fixing it but I guess I forgot about it. Thanks for finding it!

PR #104, opened 2022-08-14
Save paths in human-readable format.

bckircher:
Using a human-readable format for the JSON path output has two advantages:

1. It makes it much easier for a version control system to track and highlight what has changed in a path.
2. It makes it easier to hand-edit the path (for precise control-point locations...if you're a bit AR like myself).


bckircher:
Is there something I need to setup in my GitHub account or my cloned repository to get the Format Code / Format Dart check to pass? The check itself seems to be failing to run, not that it is having a problem with my changes (not saying that there aren't formatting problems with my changes!...though I'd prefer to make sure there aren't).

mjansen4857:
I'm not really sure why it fails. It seems to fail on cloning the repo but that works fine in the unit test action so I have no idea. It's not really a huge deal since it will be formatted next time I make a pull request myself or edit the same file.

Issue #105, opened 2022-08-17
Better marker time calculation?


bckircher:
The current marker time calculation involves looping over every state of the generated trajectory to find the one that is closest in x/y position to each marker. Not only is this slow (as recognized in the code comments!), but it will produce incorrect/unpredictable results if a trajectory consists of several waypoints that result in an overlapping trajectory (sure, perhaps this is a bit of an exceptional case, but I think it becomes more realistic with the ability to use markers to trigger events along the path following and possibly an amenable autonomous game setup).

Since the markers are defines as the "percentage" of the way along a trajectory, wouldn't this be the same as the computed trajectory state at the same "percentage" of the way along the states?

Take a simple path that has just a starting and ending waypoint. The states are computed by finding the N points that are "equally spaced" along the cubic bezier (i.e. computed using equally spaced values of _t_). For a marker placed at 0.5 along that path, the same cubic bezier is used with _t_ set to 0.5 to find the marker position.

Seems like given the step value for _t_, the two states on either side of a marker can be trivially found (by direct index computation), then either the closest or a quick lerp used...for the position in the GUI or for the time in robot code.

Of course, this becomes slightly trickier for a trajectory with more than two waypoints...but not very much (the first N states correspond to the trajectory between waypoints 1 and 2, the next N states are between waypoint 2 and 3, etc.; being careful to observe the N-1 nature since the last state of the first segment is the same as the first state of the second segment).

Have I reasoned this out correctly? Or am I missing something? If I have this right, I'm happy to develop/test the code and submit a PR.

mjansen4857:
I believe that sounds correct. I forgot at the time that each segment of a path has a fixed number of states. Older versions used to split states by time so the number per segment was variable. 

A really simple solution I could think of would just take the `waypointRelativePos` of the event marker and multiply it by the number of states per segment (found by `1.0 / PathPlanner.resolution`) and either round that to get the index of the corresponding state, or lerp between the 2 closest states.

So for example, in a path with 3 waypoints (2 segments) and a marker with a position of 1.5 (halfway between the second and third waypoint):
`i = 1.5 * (1.0 / PathPlanner.resolution (default 0.004)) = 375`

Feel free to take this on if you'd like.

PR #106, opened 2022-08-26
Improved marker time calculation:

bckircher:
- Add resolution as a constant in Trajectory
- Lerp between trajectory states to find the marker time

Fixes mjansen4857/pathplanner#105

### DIFF
--- a/lib/robot_path/robot_path.dart
+++ b/lib/robot_path/robot_path.dart
@@ -175,6 +175,7 @@ class RobotPath {
 
 class EventMarker {
   double position;
+  late double timeSeconds;
   String name;
 
   EventMarker(this.position, this.name);
--- a/lib/services/generator/trajectory.dart
+++ b/lib/services/generator/trajectory.dart
@@ -72,7 +72,11 @@ class Trajectory {
       }
     }
 
-    return Trajectory.joinTrajectories(trajectories);
+    Trajectory trajectory = Trajectory.joinTrajectories(trajectories);
+
+    calculateMarkerTimes(trajectory, path.markers);
+
+    return trajectory;
   }
 
   static Future<Trajectory> generateSingleTrajectory(List<Waypoint> pathPoints,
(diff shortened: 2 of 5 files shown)


mjansen4857:
I don't think the addition to path_painter_util.dart is necessary. Marker time is not used to paint the path and marker info is not output to pre-generated paths.

bckircher:
Correct, it is not used at present. It will be used by something I'm working on though. My thinking was to put all the marker time computation changes into a single PR, but I can pull the path_painter_util.dart part out and put that into the other PR if you'd prefer.

And path_painter_util.dart doesn't feel like the right place for it anyway...but trajectory.dart doesn't match the Java/C++ equivalents (structurally anyway).

mjansen4857:
Then you’d want to add the time field to the EventMarker class in robot_path.dart and calculate all the markers whenever the path is generated. It’s probably best to move that method to a private method of RobotPath then since it doesn’t really have anything to do with the path painter. What’s do you plan on doing with it? Adding it to pre-generated files?

bckircher:
OK, makes sense. Let me go away and re-formulate this and re-submit it.

Here's what I'm working on:

image

The marker times are needed to place them in the correct position across the top of this graph (in this example, I have markers at 0, 0.1, 0.2, ..., 0.9, 1.0).

Of course, it displays the appropriate traces for holonomic vs. non-holonomic modes.

mjansen4857:
Looks awesome. I was thinking of doing something similar to this at some point so that's a huge help. I know you're still working on it so some of this might already be done but I have a few suggestions I can think of at the moment:
- Always show the generator settings card in this editor mode so settings settings can be changed without switching mode (like how it works with the preview mode)
- Click on the labels (velocity, heading, etc.) to toggle showing/hiding them on the graph to make it easier to look at specific values (if you're using a package for the graph it might not let you do this easily so not a huge deal)
- Maybe a subtle dotted line extending down from each marker (again, not a huge deal if the package makes this difficult)

Overall, looks great. Thanks!

bckircher:
Glad you like it! I'm a noob w.r.t Dart (C++ at work, Java with the team), so this is entirely hand-written (for better or worse!). So:

- Adding the generator settings card...should be easy (just lift it from the preview editor).
- Click on the labels...completely agree and already in the process of making that happen!
- Dotted line on the marker...great idea!

bckircher:
OK, the marker time calculation has been moved out of path_painter_util.dart and into trajectory.dart (reasonably in line with what is done in the Java/C++ version). The marker time is stored EventMarker (in robot_path.dart) using a "hidden" field (doesn't show up in the constructor, Json, toString, etc., since it is a derived/computed value).

Happy to make further changes if something is still amiss!

Issue #108, opened 2022-09-04
Add the ability to plot the velocity/acceleration/etc. for the generated trajectory


bckircher:
In a new editor, plot the appropriate set of traces (velocity, acceleration, heading, angular velocity, curvature for non-holonomic robots, velocity, acceleration, rotation, and angular velocity for holonomic robots) so that the commanded behavior of the robot can be viewed. The editor would look something like the following:

<img width="1278" alt="image" src="https://user-images.githubusercontent.com/36779728/188315332-22b1ae22-79f1-4157-b914-0b75405e2113.png">

Traces could be turned on or off by clicking on them in the legend at the bottom, allow certain traces to be isolated (or uncovered if traces overlapped at some point of interest).

bckircher:
The initial version of this is at https://github.com/bckircher/pathplanner/tree/graph_editor. @mjansen4857  when you have a chance, try it out and let me know what you think. Both UX and code structure. Once we're both happy with it, I'll create a PR.

Note there is one somewhat unrelated change in here right now, which I'm happy to pull out separately if that is better. The click radius for markers was reduced from 40 to 15 (which is about the size of the marker itself). At 40, it was really hard to select markers that were located close to each other (it was easy to "hide" a marker behind the large click radius of another marker). It's set to 15 in the new graph editor, and I updated it to 15 in the marker editor to be consistent.

The graph editor code feels a bit icky to me, but I'm not sure if that is the code itself, my still relative unfamiliarity with Dart/Flutter, or a combination of both!

mjansen4857:
Haven't gotten to play with it yet, but I just merged a PR which updates flutter to 3.3.1 and changes the linter configuration to find a lot more best practice stuff. I had to fix 300+ issues so there will probably be some in this branch if you copied anything. All you should have to do is make sure you're updated to 3.3.1 and then go through the "Problems" tab in VSCode and the quick fix option it suggests should be good. If there's anything you're unsure of let me know.

I will probably make changes to this when I get around to #97 to show target vs actual values when running path following commands. If I see anything that can be improved I'll probably just do it then. There's likely a package on https://pub.dev/ that can be used to make a graph without having to paint everything manually so between now and then I'll look around for one that can match the features this has since using a full featured graph widget will be a lot easier to maintain.

bckircher:
OK, makes sense. I'll pull in the latest changes to this branch (and take care of any lint errors I'm responsible for!), then see what I can find in the flutter package repo to redo the graph.

[...]

PR #121, opened 2022-10-16
graph trajectory using fl_chart

bckircher:
The trajectory is now graphed using fl_chart. #108

### DIFF
--- a/lib/widgets/path_editor/cards/graph_settings_card.dart
+++ b/lib/widgets/path_editor/cards/graph_settings_card.dart
@@ -5,9 +5,9 @@ import 'package:shared_preferences/shared_preferences.dart';
 
 class GraphSettingsCard extends StatefulWidget {
   final bool holonomicMode;
-  final VoidCallback onUpdate;
   final VoidCallback onToggleSampled;
   final VoidCallback onToggleMinimized;
+  final VoidCallback onShouldRedraw;
   final GlobalKey stackKey;
   final SharedPreferences prefs;
   final bool isSampled;
@@ -16,9 +16,9 @@ class GraphSettingsCard extends StatefulWidget {
   const GraphSettingsCard(
       {required this.stackKey,
       required this.holonomicMode,
-      required this.onUpdate,
       required this.onToggleSampled,
       required this.onToggleMinimized,
+      required this.onShouldRedraw,
       required this.prefs,
       required this.isSampled,
       required this.cardMinimized,
@@ -29,6 +29,25 @@ class GraphSettingsCard extends StatefulWidget {
 }
 
 class _GraphSettingsCardState extends State<GraphSettingsCard> {
+  late bool _showVelocity;
+  late bool _showAccel;
+  late bool _showHeading;
+  late bool _showAngularVelocity;
+  late bool _showCurvature;
+  late bool _showRotation;
+
+  @override
+  void initState() {
+    super.initState();
+
+    _showVelocity = widget.prefs.getBool(GraphEditor.prefShowVelocity) ?? true;
+    _showAccel = widget.prefs.getBool(GraphEditor.prefShowAccel) ?? true;
+    _showHeading = widget.prefs.getBool(GraphEditor.prefShowHeading) ?? true;
+    _showAngularVelocity = widget.prefs.getBool(GraphEditor.prefShowAngularVelocity) ?? true;
+    _showCurvature = widget.prefs.getBool(GraphEditor.prefShowCurvature) ?? true;
+    _showRotation = widget.prefs.getBool(GraphEditor.prefShowRotation) ?? true;
+  }
+
   @override
   Widget build(BuildContext context) {
     return DraggableCard(
@@ -140,15 +159,16 @@ class _GraphSettingsCardState extends State<GraphSettingsCard> {
       mainAxisSize: MainAxisSize.max,
       children: [
         Checkbox(
-          value: widget.prefs.getBool(GraphEditor.prefShowVelocity),
+          value: _showVelocity,
           activeColor: GraphEditor.colorVelocity,
           checkColor: GraphEditor.colorVelocity,
           side: BorderSide(color: GraphEditor.colorVelocity, width: 2),
           onChanged: (val) {
-            bool state =
-                widget.prefs.getBool(GraphEditor.prefShowVelocity) ?? true;
-            widget.prefs.setBool(GraphEditor.prefShowVelocity, !state);
-            widget.onUpdate();
+            widget.prefs.setBool(GraphEditor.prefShowVelocity, val!);
+            setState(() {
+              _showVelocity = val;
+            });
+            widget.onShouldRedraw();
           },
         ),
         Text(
@@ -157,15 +177,16 @@ class _GraphSettingsCardState extends State<GraphSettingsCard> {
         ),
         const SizedBox(width: 12),
         Checkbox(
-          value: widget.prefs.getBool(GraphEditor.prefShowAccel),
+          value: _showAccel,
           activeColor: GraphEditor.colorAccel,
           checkColor: GraphEditor.colorAccel,
           side: BorderSide(color: GraphEditor.colorAccel, width: 2),
           onChanged: (val) {
-            bool state =
-                widget.prefs.getBool(GraphEditor.prefShowAccel) ?? true;
-            widget.prefs.setBool(GraphEditor.prefShowAccel, !state);
-            widget.onUpdate();
+            widget.prefs.setBool(GraphEditor.prefShowAccel, val!);
+            setState(() {
+              _showAccel = val;
+            });
+            widget.onShouldRedraw();
           },
         ),
         Text(
@@ -184,15 +205,16 @@ class _GraphSettingsCardState extends State<GraphSettingsCard> {
       mainAxisSize: MainAxisSize.max,
       children: [
         Checkbox(
-          value: widget.prefs.getBool(GraphEditor.prefShowHeading),
+          value: _showHeading,
           activeColor: GraphEditor.colorHeading,
           checkColor: GraphEditor.colorHeading,
           side: BorderSide(color: GraphEditor.colorHeading, width: 2),
           onChanged: (val) {
-            bool state =
-                widget.prefs.getBool(GraphEditor.prefShowHeading) ?? true;
-            widget.prefs.setBool(GraphEditor.prefShowHeading, !state);
-            widget.onUpdate();
+            widget.prefs.setBool(GraphEditor.prefShowHeading, val!);
+            setState(() {
+              _showHeading = val;
+            });
+            widget.onShouldRedraw();
           },
         ),
         Text(
@@ -201,16 +223,16 @@ class _GraphSettingsCardState extends State<GraphSettingsCard> {
         ),
         const SizedBox(width: 12),
         Checkbox(
-          value: widget.prefs.getBool(GraphEditor.prefShowAngularVelocity),
+          value: _showAngularVelocity,
           activeColor: GraphEditor.colorAngularVelocity,
           checkColor: GraphEditor.colorAngularVelocity,
           side: BorderSide(color: GraphEditor.colorAngularVelocity, width: 2),
           onChanged: (val) {
-            bool state =
-                widget.prefs.getBool(GraphEditor.prefShowAngularVelocity) ??
-                    true;
-            widget.prefs.setBool(GraphEditor.prefShowAngularVelocity, !state);
-            widget.onUpdate();
+            widget.prefs.setBool(GraphEditor.prefShowAngularVelocity, val!);
+            setState(() {
+              _showAngularVelocity = val;
+            });
+            widget.onShouldRedraw();
           },
         ),
         Text(
@@ -229,15 +251,16 @@ class _GraphSettingsCardState extends State<GraphSettingsCard> {
       mainAxisSize: MainAxisSize.max,
       children: [
         Checkbox(
-          value: widget.prefs.getBool(GraphEditor.prefShowCurvature),
+          value: _showCurvature,
           activeColor: GraphEditor.colorCurvature,
           checkColor: GraphEditor.colorCurvature,
           side: BorderSide(color: GraphEditor.colorCurvature, width: 2),
           onChanged: (val) {
-            bool state =
-                widget.prefs.getBool(GraphEditor.prefShowCurvature) ?? true;
-            widget.prefs.setBool(GraphEditor.prefShowCurvature, !state);
-            widget.onUpdate();
+            widget.prefs.setBool(GraphEditor.prefShowCurvature, val!);
+            setState(() {
+              _showCurvature = val;
+              widget.onShouldRedraw();
+            });
           },
         ),
         Text(
@@ -256,15 +279,16 @@ class _GraphSettingsCardState extends State<GraphSettingsCard> {
       mainAxisSize: MainAxisSize.max,
       children: [
         Checkbox(
-          value: widget.prefs.getBool(GraphEditor.prefShowRotation),
+          value: _showRotation,
           activeColor: GraphEditor.colorRotation,
           checkColor: GraphEditor.colorRotation,
           side: BorderSide(color: GraphEditor.colorRotation, width: 2),
           onChanged: (val) {
-            bool state =
-                widget.prefs.getBool(GraphEditor.prefShowRotation) ?? true;
-            widget.prefs.setBool(GraphEditor.prefShowRotation, !state);
-            widget.onUpdate();
+            widget.prefs.setBool(GraphEditor.prefShowRotation, val!);
+            setState(() {
+              _showRotation = val;
+            });
+            widget.onShouldRedraw();
           },
         ),
         Text(
@@ -273,16 +297,16 @@ class _GraphSettingsCardState extends State<GraphSettingsCard> {
         ),
         const SizedBox(width: 12),
         Checkbox(
-          value: widget.prefs.getBool(GraphEditor.prefShowAngularVelocity),
+          value: _showAngularVelocity,
           activeColor: GraphEditor.colorAngularVelocity,
           checkColor: GraphEditor.colorAngularVelocity,
           side: BorderSide(color: GraphEditor.colorAngularVelocity, width: 2),
           onChanged: (val) {
-            bool state =
-                widget.prefs.getBool(GraphEditor.prefShowAngularVelocity) ??
-                    true;
-            widget.prefs.setBool(GraphEditor.prefShowAngularVelocity, !state);
-            widget.onUpdate();
+            widget.prefs.setBool(GraphEditor.prefShowAngularVelocity, val!);
+            setState(() {
+              _showAngularVelocity = val;
+            });
+            widget.onShouldRedraw();
           },
         ),
         Text(
--- a/lib/widgets/path_editor/cards/graph_settings_card.dart
+++ b/lib/widgets/path_editor/cards/graph_settings_card.dart
@@ -43,8 +43,10 @@ class _GraphSettingsCardState extends State<GraphSettingsCard> {
     _showVelocity = widget.prefs.getBool(GraphEditor.prefShowVelocity) ?? true;
     _showAccel = widget.prefs.getBool(GraphEditor.prefShowAccel) ?? true;
     _showHeading = widget.prefs.getBool(GraphEditor.prefShowHeading) ?? true;
-    _showAngularVelocity = widget.prefs.getBool(GraphEditor.prefShowAngularVelocity) ?? true;
-    _showCurvature = widget.prefs.getBool(GraphEditor.prefShowCurvature) ?? true;
+    _showAngularVelocity =
+        widget.prefs.getBool(GraphEditor.prefShowAngularVelocity) ?? true;
+    _showCurvature =
+        widget.prefs.getBool(GraphEditor.prefShowCurvature) ?? true;
     _showRotation = widget.prefs.getBool(GraphEditor.prefShowRotation) ?? true;
   }
 
--- a/lib/widgets/path_editor/cards/graph_settings_card.dart
+++ b/lib/widgets/path_editor/cards/graph_settings_card.dart
@@ -1,0 +1,295 @@
+import 'package:flutter/material.dart';
+import 'package:pathplanner/widgets/draggable_card.dart';
+import 'package:pathplanner/widgets/path_editor/editors/graph_editor.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+class GraphSettingsCard extends StatefulWidget {
+  final bool holonomicMode;
+  final VoidCallback onUpdate;
+  final VoidCallback onToggleSampled;
+  final VoidCallback onToggleMinimized;
+  final GlobalKey stackKey;
+  final SharedPreferences prefs;
+  final bool isSampled;
+  final bool cardMinimized;
+
+  const GraphSettingsCard(
+      {required this.stackKey,
+      required this.holonomicMode,
+      required this.onUpdate,
+      required this.onToggleSampled,
+      required this.onToggleMinimized,
+      required this.prefs,
+      required this.isSampled,
+      required this.cardMinimized,
+      super.key});
+
+  @override
+  State<GraphSettingsCard> createState() => _GraphSettingsCardState();
+}
+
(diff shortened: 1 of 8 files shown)


lib/widgets/path_editor/cards/marker_card.dart
```diff
@@ -1,12 +1,14 @@
...
 import 'package:pathplanner/robot_path/robot_path.dart';
+import 'package:pathplanner/services/generator/trajectory.dart';
 import 'package:pathplanner/widgets/custom_input_slider.dart';
 import 'package:pathplanner/widgets/draggable_card.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
 class MarkerCard extends StatefulWidget {
   final GlobalKey stackKey;
   final SharedPreferences prefs;
+  final RobotPath? path;
```
mjansen4857, line 11: This doesn't need to be nullable since the path in marker_editor cannot be null.

```suggestion
  final RobotPath path;
```

lib/widgets/path_editor/cards/marker_card.dart
```diff
@@ -152,6 +155,8 @@ class _MarkerCardState extends State<MarkerCard> {
             if (widget.marker != null) {
               _oldMarker ??= widget.marker!.clone();
               widget.marker!.position = value;
+              Trajectory.calculateMarkerTime(
+                  widget.path!.generatedTrajectory, widget.marker!);
```
mjansen4857, line 159:
```suggestion
              Trajectory.calculateMarkerTime(
                  widget.path.generatedTrajectory, widget.marker!);
```

lib/widgets/path_editor/editors/graph_editor.dart
```diff
@@ -0,0 +1,427 @@
...
+
+    return SideTitleWidget(
+      axisSide: meta.axisSide,
+      space: 10,
+      child: Text(value.toStringAsFixed(1), style: style),
+    );
+  }
+}
+
+class GraphEditor extends StatefulWidget {
```
mjansen4857, line 330: I'd move this class above the `_LineChart` class so the main widget class for this file is at the top

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+      required this.prefs,
+      required this.isSampled,
+      required this.cardMinimized,
+      super.key});
+
+  @override
+  State<GraphSettingsCard> createState() => _GraphSettingsCardState();
+}
+
+class _GraphSettingsCardState extends State<GraphSettingsCard> {
```
mjansen4857, line 31: You should use booleans to represent the value of the checkboxes here. This way the app only has to rebuild the changed checkbox instead of the entire card with the UniqueKey stuff you're doing in graph_editor. This is done in generator_settings_card to make sure that the path runtime gets updated, but its not needed here. It also prevents crashes from sending giving a null value to the checkboxes due to the default values.

```suggestion
class _GraphSettingsCardState extends State<GraphSettingsCard> {
  late bool _showVelocity;
  late bool _showAccel;
  late bool _showHeading;
  late bool _showAngularVelocity;
  late bool _showCurvature;
  late bool _showRotation;
  
  @override
  void initState() {
    super.initState();
    
    _showVelocity = widget.prefs.getBool(GraphEditor.prefShowVelocity) ?? true;
    _showAcce = widget.prefs.getBool(GraphEditor.prefShowAccel) ?? true;
    _showHeading = widget.prefs.getBool(GraphEditor.prefShowheading) ?? true;
    _showAngularVelocity = widget.prefs.getBool(GraphEditor.prefShowAngularVelocity) ?? true;
    _showCurvature = widget.prefs.getBool(GraphEditor.prefShowCurvature) ?? true;
    _showRotation = widget.prefs.getBool(GraphEditor.prefShowRotation) ?? true;
  }
  
```

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
+import 'package:flutter/material.dart';
+import 'package:pathplanner/widgets/draggable_card.dart';
+import 'package:pathplanner/widgets/path_editor/editors/graph_editor.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+class GraphSettingsCard extends StatefulWidget {
+  final bool holonomicMode;
+  final VoidCallback onUpdate;
```
mjansen4857, line 8: Not needed with changes below

```suggestion
```

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+  final VoidCallback onToggleMinimized;
+  final GlobalKey stackKey;
+  final SharedPreferences prefs;
+  final bool isSampled;
+  final bool cardMinimized;
+
+  const GraphSettingsCard(
+      {required this.stackKey,
+      required this.holonomicMode,
+      required this.onUpdate,
```
mjansen4857, line 19: Not needed with changes below

```suggestion
```

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+        Checkbox(
+          value: widget.prefs.getBool(GraphEditor.prefShowVelocity),
+          activeColor: GraphEditor.colorVelocity,
+          checkColor: GraphEditor.colorVelocity,
+          side: BorderSide(color: GraphEditor.colorVelocity, width: 2),
+          onChanged: (val) {
+            bool state =
+                widget.prefs.getBool(GraphEditor.prefShowVelocity) ?? true;
+            widget.prefs.setBool(GraphEditor.prefShowVelocity, !state);
+            widget.onUpdate();
```
mjansen4857, line 151:
```suggestion
            widget.prefs.setBool(GraphEditor.prefShowVelocity, val);
            setState({
              _showVelocity = val;
            });
```

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+        Checkbox(
+          value: widget.prefs.getBool(GraphEditor.prefShowAccel),
+          activeColor: GraphEditor.colorAccel,
+          checkColor: GraphEditor.colorAccel,
+          side: BorderSide(color: GraphEditor.colorAccel, width: 2),
+          onChanged: (val) {
+            bool state =
+                widget.prefs.getBool(GraphEditor.prefShowAccel) ?? true;
+            widget.prefs.setBool(GraphEditor.prefShowAccel, !state);
+            widget.onUpdate();
```
mjansen4857, line 168:
```suggestion
            widget.prefs.setBool(GraphEditor.prefShowAccel, val);
            setState({
              _showAccel = val;
            });
```

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+        Checkbox(
+          value: widget.prefs.getBool(GraphEditor.prefShowHeading),
+          activeColor: GraphEditor.colorHeading,
+          checkColor: GraphEditor.colorHeading,
+          side: BorderSide(color: GraphEditor.colorHeading, width: 2),
+          onChanged: (val) {
+            bool state =
+                widget.prefs.getBool(GraphEditor.prefShowHeading) ?? true;
+            widget.prefs.setBool(GraphEditor.prefShowHeading, !state);
+            widget.onUpdate();
```
mjansen4857, line 195:
```suggestion
            widget.prefs.setBool(GraphEditor.prefShowHeading, val);
            setState({
              _showHeading = val;
            });
```

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+          value: widget.prefs.getBool(GraphEditor.prefShowAngularVelocity),
+          activeColor: GraphEditor.colorAngularVelocity,
+          checkColor: GraphEditor.colorAngularVelocity,
+          side: BorderSide(color: GraphEditor.colorAngularVelocity, width: 2),
+          onChanged: (val) {
+            bool state =
+                widget.prefs.getBool(GraphEditor.prefShowAngularVelocity) ??
+                    true;
+            widget.prefs.setBool(GraphEditor.prefShowAngularVelocity, !state);
+            widget.onUpdate();
```
mjansen4857, line 213:
```suggestion
            widget.prefs.setBool(GraphEditor.prefShowAngularVelocity, val);
            setState({
              _showAngularVelocity = val;
            });
```

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+        Checkbox(
+          value: widget.prefs.getBool(GraphEditor.prefShowCurvature),
+          activeColor: GraphEditor.colorCurvature,
+          checkColor: GraphEditor.colorCurvature,
+          side: BorderSide(color: GraphEditor.colorCurvature, width: 2),
+          onChanged: (val) {
+            bool state =
+                widget.prefs.getBool(GraphEditor.prefShowCurvature) ?? true;
+            widget.prefs.setBool(GraphEditor.prefShowCurvature, !state);
+            widget.onUpdate();
```
mjansen4857, line 240:
```suggestion
            widget.prefs.setBool(GraphEditor.prefShowCurvature, val);
            setState({
              _showCurvature = val;
            });
```

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+        Checkbox(
+          value: widget.prefs.getBool(GraphEditor.prefShowRotation),
+          activeColor: GraphEditor.colorRotation,
+          checkColor: GraphEditor.colorRotation,
+          side: BorderSide(color: GraphEditor.colorRotation, width: 2),
+          onChanged: (val) {
+            bool state =
+                widget.prefs.getBool(GraphEditor.prefShowRotation) ?? true;
+            widget.prefs.setBool(GraphEditor.prefShowRotation, !state);
+            widget.onUpdate();
```
mjansen4857, line 267:
```suggestion
            widget.prefs.setBool(GraphEditor.prefShowRotation, val);
            setState({
              _showRotation = val;
            });
```

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+          value: widget.prefs.getBool(GraphEditor.prefShowAngularVelocity),
+          activeColor: GraphEditor.colorAngularVelocity,
+          checkColor: GraphEditor.colorAngularVelocity,
+          side: BorderSide(color: GraphEditor.colorAngularVelocity, width: 2),
+          onChanged: (val) {
+            bool state =
+                widget.prefs.getBool(GraphEditor.prefShowAngularVelocity) ??
+                    true;
+            widget.prefs.setBool(GraphEditor.prefShowAngularVelocity, !state);
+            widget.onUpdate();
```
mjansen4857, line 285:
```suggestion
            widget.prefs.setBool(GraphEditor.prefShowAngularVelocityl, val);
            setState({
              _showAngularVelocity = val;
            });
```

lib/widgets/path_editor/editors/graph_editor.dart
```diff
@@ -0,0 +1,427 @@
...
+      required this.prefs,
+      super.key});
+
+  @override
+  State<GraphEditor> createState() => _GraphEditorState();
+}
+
+class _GraphEditorState extends State<GraphEditor> {
+  final GlobalKey _key = GlobalKey();
+  UniqueKey _graphRuntimeKey = UniqueKey();
```
mjansen4857, line 361: Not needed with the changes to graph_settings_card

```suggestion
```

lib/widgets/path_editor/editors/graph_editor.dart
```diff
@@ -0,0 +1,427 @@
...
+  bool isSampled = false;
+  bool cardMinimized = false;
```
mjansen4857, line 363: Prefix with `_` to keep these private unless they're accessed from another class. Not really a huge deal its just a bit more "correct"

```suggestion
  bool _isSampled = false;
  bool _cardMinimized = false;
```

lib/widgets/path_editor/editors/graph_editor.dart
```diff
@@ -0,0 +1,427 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Stack(
+      key: _key,
+      children: [
+        Padding(
+          padding: const EdgeInsets.fromLTRB(48, 48, 48, 72),
+          child: _LineChart(
+              isSampled: isSampled,
```
mjansen4857, line 373:
```suggestion
              isSampled: _isSampled,
```

lib/widgets/path_editor/editors/graph_editor.dart
```diff
@@ -0,0 +1,427 @@
...
+        _buildGraphSettingsCard(),
+        _buildGeneratorSettingsCard(),
+      ],
+    );
+  }
+
+  Widget _buildGraphSettingsCard() {
+    return GraphSettingsCard(
+        stackKey: _key,
+        key: _graphRuntimeKey,
```
mjansen4857, line 387: Not needed with changes to graph_settings_card

```suggestion
```

lib/widgets/path_editor/editors/graph_editor.dart
```diff
@@ -0,0 +1,427 @@
...
+        holonomicMode: widget.holonomicMode,
+        onUpdate: () async {
+          setState(() {
+            // Force rebuild card to update runtime
+            _graphRuntimeKey = UniqueKey();
+          });
+        },
```
mjansen4857, line 394: Not needed with changes to graph_settings_card

```suggestion
```

lib/widgets/path_editor/editors/graph_editor.dart
```diff
@@ -0,0 +1,427 @@
...
+        onToggleSampled: () async {
+          setState(() {
+            isSampled = !isSampled;
+          });
+        },
+        onToggleMinimized: () async {
+          setState(() {
+            cardMinimized = !cardMinimized;
+          });
+        },
```
mjansen4857, line 404: Doesn't need to be async since you're not running any async code

```suggestion
        onToggleMinimized: () {
          setState(() {
            _cardMinimized = !_cardMinimized;
          });
        },
```

lib/widgets/path_editor/editors/graph_editor.dart
```diff
@@ -0,0 +1,427 @@
...
+        prefs: widget.prefs,
+        isSampled: isSampled,
+        cardMinimized: cardMinimized);
```
mjansen4857, line 407:
```suggestion
        _isSampled: _isSampled,
        _cardMinimized: _cardMinimized,
      );
```

mjansen4857 (changes requested):
App crashed when I tried to run it so I requested some changes that should fix that + make some improvements. I may have made typos or messed something up I didn't see so hopefully I didn't break anything.

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -0,0 +1,295 @@
...
+
+  Widget _buildCurvatureRow(BuildContext context) {
+    ColorScheme colorScheme = Theme.of(context).colorScheme;
+
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.center,
+      mainAxisSize: MainAxisSize.max,
+      children: [
+        Checkbox(
+          value: widget.prefs.getBool(GraphEditor.prefShowCurvature),
```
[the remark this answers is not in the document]
bckircher: OK, I can see how this might make it crash for you. I have those preferences set through numerous iterations to get from nothing to what you see now...so I've ended up with something that "works for me" as a result. I've looked around (for maybe 10 minutes...with no luck) trying to figure out how to wipe out all the PathPlanner preferences from my computer so I can run it from a fresh system. I was hoping to be able to reproduce the crash you see, and therefore confirm that the next set of changes will fix that crash. What you propose (and I've done) _seems_ like it should do the trick. Knowing instead of guessing would be so much better! ;-)

lib/widgets/path_editor/editors/graph_editor.dart
mjansen4857: The only thing that you are passing _graphRuntimeKey to that I can see is the GraphSettingsCard widget so I don't think this is actually causing the graph itself to rebuild. I'm not sure though.

lib/widgets/path_editor/cards/graph_settings_card.dart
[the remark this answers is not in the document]
mjansen4857: Well the cause is from not having those preferences saved previously so the default values in initState will fix it but if you wanted to make extra sure you can delete shared_preferences.json in `%AppData%\com.mjansen4857\pathplanner` assuming you are on windows. If you're on mac its more trouble than its worth.

lib/widgets/path_editor/editors/graph_editor.dart
bckircher: The graph isn't rebuilt without it! Not sure if it is because of the _graphRuntimeKey or the setState() by itself...I'm still a bit unclear on the Flutter event model.

I followed the pattern used on other cards/editors, but may have mis-applied it a bit. I got rid of _graphRuntimeKey and have an empty lambda in setState() and it redraws the graph. I'm happy with that if you are.

mjansen4857: I'm wondering if that's because of the GlobalKey you have. Not sure how that works. Does it still rebuild the graph if you use a UniqueKey for _key and use `_key = UniqueKey()` in set state? If that works it would be a bit more clear that rebuilding the graph is what its doing.

lib/widgets/path_editor/cards/graph_settings_card.dart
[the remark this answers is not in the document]
mjansen4857: You could do it, you just have to delete preferences individually from the command line. There might be a way to delete all of them but I'm unsure an example would be: `defaults delete com.mjansen4857.pathplanner flutter.showVelocity`

Although after looking at the preference keys you have set, I'd probably change them to have a prefix of "graph" or something similar so they aren't so generic. Changing the names would have the same effect of deleting all of them.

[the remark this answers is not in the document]
bckircher: Good general suggestion on the setting prefix. After adding it, the app crashed for me as well. After the initState() fixes, it no longer crashes. Yeah!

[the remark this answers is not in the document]
bckircher: Oh, `defaults delete com.mjansen4857.pathplanner` will delete all the settings for PathPlanner. Simple!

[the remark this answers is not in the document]
mjansen4857: Ah, ok. Didn't want to try that myself incase it messes something else up in the defaults that sharedpreferences doesn't manage.

lib/widgets/path_editor/editors/graph_editor.dart
mjansen4857: Didn't notice that. It should be fine but what you could do instead is create a new field for the UniqueKey and only pass that to the _LineChart widget.

bckircher:
OK, I've updated most of what you suggested above. Still up in the air is the onShouldRedraw() thing. I'll think about it and look at it again tomorrow. I'm open to your thoughts and suggestions of course.

lib/widgets/path_editor/cards/graph_settings_card.dart
```diff
@@ -229,15 +253,16 @@ class _GraphSettingsCardState extends State<GraphSettingsCard> {
...
           onChanged: (val) {
-            bool state =
-                widget.prefs.getBool(GraphEditor.prefShowCurvature) ?? true;
-            widget.prefs.setBool(GraphEditor.prefShowCurvature, !state);
-            widget.onUpdate();
+            widget.prefs.setBool(GraphEditor.prefShowCurvature, val!);
+            setState(() {
+              _showCurvature = val;
+              widget.onShouldRedraw();
+            });
```
mjansen4857, line 265: Being a bit pedantic here but I'd move the onShouldRedraw call out of setState to match all the other calls lol

```suggestion
            setState(() {
              _showCurvature = val;
            });
            widget.onShouldRedraw();
```

lib/widgets/path_editor/editors/graph_editor.dart
bckircher: Sounds reasonable. I'll look into that.

lib/widgets/path_editor/cards/graph_settings_card.dart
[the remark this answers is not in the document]
mjansen4857: Ok I'll look into that now. I did that so long ago I don't remember either. That might explain a crash someone else was having earlier... oops.

mjansen4857 (commented):
Made one more little request that doesn't really matter but overall looks good. I'll run it and play around with the graph a bit to see if I can find anything else before merging.

mjansen4857:
After playing around with it, it seems pretty nice. There's a couple little improvements I would make but I can do that after its merged + I'm not really sure about them yet. Let me know when you're happy with it and I can merge.

lib/widgets/path_editor/cards/graph_settings_card.dart
bckircher: Oops... <looks sheepishly away>

lib/widgets/path_editor/editors/graph_editor.dart
bckircher: So, the _LineChart widget is a StatelessWidget, meaning that the UniqueKey passed to it with this approach would not change (if I'm understanding Dart correctly, it is pass by value only, so it couldn't change; and the key in _LineChart would have to be final since it is a StatelessWidget, correct?).

In the end, I think this ends up just being syntactic sugar. This is an odd case in that the graph needs to be redrawn based on "state" stored in the preferences that is updated by the card...so there is no state update required in the GraphEditor or _LineChart widgets.

I'm leaning towards an empty lambda with a comment that describes what is happening. As opposed to trying to imply it with a meaningless "state" change and hoping that someone understands it a couple of years from now. Thoughts?

bckircher:
OK, I believe that I've address all your concerns and feedback (thanks!). I ended up going with the empty setState() and a comment describing what it does. If a better solution is found at some point, it can always be changed.

I'm happy with this PR if you are, and am ready for it to be merged.

mjansen4857:
Just a heads up so you know why I made a few changes:

I did a bunch of re-organizing in graph_editor to optimize it a bit. Previously, you were creating all of the graph data in the build method of the _LineChart widget. This is a bit unoptimized since it will loop through the whole path every time the graph is rebuilt. So, I moved stuff around so now the FlSpot lists are only created in initState, or when generator settings are changed. I also made it so that graph_editor stores these lists for every value, regardless of whether they are enabled in the graph settings card or not. This makes it so when one of the checkboxes in the card is changed, the graph_editor can show/hide that value instead of rebuilding all of the data and the whole graph. Now, only parts of the graph that change will need to rebuild, instead of the whole thing.

Hopefully that explanation makes sense lol. Basically, flutter is designed to only rebuild widgets that have been changed, so if you can avoid changing things that don't need to be changed (i.e. creating new FlSpot lists every time build is called) it won't need to rebuild as much stuff since it knows the lists haven't changed.

bckircher:
Awesome, not a problem! I hadn't given any thought to the efficiency side of things (maybe because a computer is "infinitely faster" than the embedded systems I program at work, and I needed a break from obsessing over efficiency!). Plus, I'm still learning this Dart/Flutter thing, so the more examples I see of better ways to do things, especially looking at the efficiency changes you've made to my code, will only help me.

Thanks for a great tool, and for allowing me to help make it better!

mjansen4857:
Yeah definitely, that’s why I wanted to let you know. It won’t really make a noticeable difference to most people but I like to try to optimize everything to some degree considering some of the awful laptops some teams have to use lol. Thanks for the contribution, really appreciate it. Already helped me fix a generation bug I didn’t know about.

bckircher:
LOL! I knew about that generation "strangeness", and had the same fix that you committed. But was wanting to get there via an easy way to look at the data that was generated. So, it all worked out!

Issue #147, opened 2022-11-14
Many, if not most, of the C++ pathplannerlib functions could be declared const, but they are not


MrRedness:
I would like to try implementing this tomorrow, provided I get to it, so please don't feel rushed to do it (I've seen how fast you've completed my and others' requests). It would help with writing more resource-efficient code, at least in my team's codebase.

PR #150, opened 2022-11-14
Const & constexpr, refactoring, format

MrRedness:
Follows #147


MrRedness:
I can test on a robot tomorrow

MrRedness:
Confirmed compiles in robot project

mjansen4857:
Could you try to undo the formatting changes? They seem to be your personal formatting settings as the WPILib formatting and the formatting I use use opening brace on same line style.

MrRedness:
Sure. I'll also check on that build failure. It works fine for me, but it's probably because I'm using the 2023 compiler and libraries.

mjansen4857:
Actions is using the 2023 wpilib docker container so not sure why it's different.

MrRedness:
I don't know either, but for know I'll just remove that as being constexpr

MrRedness:
It builds fine on 2022
Also, I check my formatter settings, and it appears I'm using the defaults (I just installed the new version of VSCode last week). Perhaps you're using the java formatter?

MrRedness:
Fixed formatting

MrRedness:
Confirmed compiles in robot project

mjansen4857:
Out of curiosity, are you working on windows?

After pulling changes, I was unable to build pplib because a few of the methods in GeometryUtil could not result in a consexpr. I looked into it a bit and it looks like the cos, isfinite, and isnan methods were not made constexpr until C++ 23, but 2023 wpilib uses C++ 20. I was able to fix the issues and build, but it's weird that your build and CI doesn't fail. I'd expect the 2023 docker containers to be setup up with C++ 20 and not build with 23.

MrRedness:
I'm working on windows because I could never get the WPILIB build system set-up on Linux, but then again it's been a year since I tried. The standard library versions should be the same, regardless of any compiler differences. Very strange behavior.

-------- Original Message --------

PR #151, opened 2022-11-17
More optimizations for C++

MrRedness:
Includes optimizing movement of vectors, as well as objects into vectors. Also includes most variables becoming const (compiler can optimize these variables), and a few general optimizations. Let me know if there's anything about the previous format that you'd prefer to retain.

### DIFF
--- a/pathplannerlib/src/main/native/cpp/pathplanner/lib/PathPlanner.cpp
+++ b/pathplannerlib/src/main/native/cpp/pathplanner/lib/PathPlanner.cpp
@@ -261,11 +261,11 @@ std::vector<PathPlannerTrajectory::EventMarker> PathPlanner::getMarkersFromJson(
 			std::vector < std::string > names;
 			if (marker.find("names") != marker.end()) {
 				for (wpi::json::const_reference name : marker.at("names")) {
-					names.emplace_back(static_cast<std::string const&>(name));
+					names.push_back(name);
 				}
 			} else {
 				// Handle transition from one-event markers to multi-event markers. Remove next season
-				names.emplace_back(marker.at("name"));
+				names.push_back(marker.at("name"));
 			}
 			markers.emplace_back(
 					PathPlannerTrajectory::EventMarker(std::move(names),
--- a/pathplannerlib/src/main/native/cpp/pathplanner/lib/PathPlanner.cpp
+++ b/pathplannerlib/src/main/native/cpp/pathplanner/lib/PathPlanner.cpp
@@ -14,8 +14,8 @@ using namespace pathplanner;
 double PathPlanner::resolution = 0.004;
 
 PathPlannerTrajectory PathPlanner::loadPath(std::string const &name,
-		PathConstraints const &constraints, bool reversed) {
-	std::string filePath = frc::filesystem::GetDeployDirectory()
+		PathConstraints const constraints, bool const reversed) {
+	std::string const filePath = frc::filesystem::GetDeployDirectory()
 			+ "/pathplanner/" + name + ".path";
 
 	std::error_code error_code;
@@ -38,14 +38,14 @@ PathPlannerTrajectory PathPlanner::loadPath(std::string const &name,
 
 std::vector<PathPlannerTrajectory> PathPlanner::loadPathGroup(
 		std::string const &name,
-		std::initializer_list<PathConstraints> const &constraints,
-		bool reversed) {
+		std::initializer_list<PathConstraints> const constraints,
+		bool const reversed) {
 	if (constraints.size() == 0) {
 		throw std::runtime_error(
 				"At least one PathConstraints is required but none were provized");
 	}
 
-	std::string filePath = frc::filesystem::GetDeployDirectory()
+	std::string const filePath = frc::filesystem::GetDeployDirectory()
 			+ "/pathplanner/" + name + ".path";
 
 	std::error_code error_code;
@@ -70,21 +70,21 @@ std::vector<PathPlannerTrajectory> PathPlanner::loadPathGroup(
 
 	std::vector < PathPlannerTrajectory::Waypoint > currentPath;
 	for (size_t i = 0; i < waypoints.size(); i++) {
-		PathPlannerTrajectory::Waypoint const &w = waypoints[i];
+		PathPlannerTrajectory::Waypoint const w = waypoints[i];
 
 		currentPath.push_back(w);
 		if (w.isStopPoint || i == waypoints.size() - 1) {
 			// Get the markers that should be part of this path and correct their positions
 			std::vector < PathPlannerTrajectory::EventMarker > currentMarkers;
-			for (PathPlannerTrajectory::EventMarker marker : markers) {
+			for (PathPlannerTrajectory::EventMarker const &marker : markers) {
 				if (marker.waypointRelativePos
 						>= indexOfWaypoint(waypoints, currentPath[0])
 						&& marker.waypointRelativePos <= i) {
-					currentMarkers.emplace_back(
-                            std::move(marker.names),
-                            marker.waypointRelativePos
-                                    - indexOfWaypoint(waypoints,
-                                            currentPath[0]));
+					currentMarkers.push_back(
+							PathPlannerTrajectory::EventMarker(marker.names,
+									marker.waypointRelativePos
+											- indexOfWaypoint(waypoints,
+													currentPath[0])));
 				}
 			}
 			splitMarkers.push_back(currentMarkers);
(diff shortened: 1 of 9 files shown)


pathplannerlib/src/main/native/include/pathplanner/lib/PathConstraints.h
```diff
@@ -3,6 +3,8 @@
 #include <units/velocity.h>
 #include <units/acceleration.h>
 
+#include <assert.h>
+
```
mjansen4857, line 7: This is unused, correct?

MrRedness: Ah, yes, sorry. I added that while testing to see which classes are trivially copyable, since those classes are better off passed by value than reference.

pathplannerlib/src/main/native/cpp/pathplanner/lib/PathPlanner.cpp
```diff
@@ -262,19 +257,19 @@ std::vector<PathPlannerTrajectory::EventMarker> PathPlanner::getMarkersFromJson(
...
-		for (wpi::json::reference marker : json.at("markers")) {
+		for (wpi::json::const_reference marker : json.at("markers")) {
 			std::vector < std::string > names;
 			if (marker.find("names") != marker.end()) {
-				for (std::string name : marker.at("names")) {
-					names.push_back(name);
+				for (wpi::json::const_reference name : marker.at("names")) {
+					names.emplace_back(name);
 				}
 			} else {
```
mjansen4857, line 266: This change is causing a build error for me on my machine. Not sure what the difference is but here is the error output:

```
PathPlanner.cpp
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\xmemory(676): error C2672: 'std::construct_at': no matching overloaded function found
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\vector(602): note: see reference to function template instantiation 'void std::_Default_allocator_traits<_Alloc>::construct<_Ty,const wpi::json::value_type&>(_Alloc &,_Objty *const ,const wpi::json::value_type &)' being compiled
        with
        [
            _Alloc=std::allocator<std::string>,
            _Ty=std::basic_string<char,std::char_traits<char>,std::allocator<char>>,
            _Objty=std::basic_string<char,std::char_traits<char>,std::allocator<char>>
        ]
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\vector(607): note: see reference to function template instantiation 'void std::_Default_allocator_traits<_Alloc>::construct<_Ty,const wpi::json::value_type&>(_Alloc &,_Objty *const ,const wpi::json::value_type &)' being compiled
        with
        [
            _Alloc=std::allocator<std::string>,
            _Ty=std::basic_string<char,std::char_traits<char>,std::allocator<char>>,
            _Objty=std::basic_string<char,std::char_traits<char>,std::allocator<char>>
        ]
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\vector(620): note: see reference to function template instantiation '_Ty &std::vector<_Ty,std::allocator<_Ty>>::_Emplace_back_with_unused_capacity<const wpi::json::value_type&>(const wpi::json::value_type &)' being compiled
        with
        [
            _Ty=std::string
        ]
C:\Dev\pathplanner\pathplannerlib\src\main\native\cpp\pathplanner\lib\PathPlanner.cpp(264): note: see reference to function template instantiation '_Ty &std::vector<_Ty,std::allocator<_Ty>>::emplace_back<const wpi::json::value_type&>(const wpi::json::value_type &)' being compiled
        with
        [
            _Ty=std::string
        ]
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\xmemory(676): error C2783: '_Ty *std::construct_at(_Ty *const ,_Types &&...) noexcept(<expr>)': could not deduce template argument for '<unnamed-symbol>'
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\xutility(134): note: see declaration of 'std::construct_at'
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\xmemory(680): error C2668: 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>::basic_string': ambiguous call to overloaded function
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\xstring(2612): note: could be 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>::basic_string(std::basic_string<char,std::char_traits<char>,std::allocator<char>> &&) noexcept'
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\xstring(2960): note: or       'std::basic_string<char,std::char_traits<char>,std::allocator<char>>::basic_string(std::initializer_list<_Elem>,const _Alloc &)'
        with
        [
            _Elem=char,
            _Alloc=std::allocator<char>
        ]
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.31.31103\include\xmemory(673): note: while trying to match the argument list '(const wpi::json::value_type)'
```

seems to be unable to convert the const_reference to a string.

mjansen4857: I updated my Visual Studio install and it still fails. I have no idea why it doesn't fail on your end.

MrRedness: Your compiler doesn't seem to enjoy solving puzzles. I'll rewrite to be a little bit more explicit.

MrRedness: Hopefully the new commit works

mjansen4857: Getting the same error unfortunately. Is your install using the MSVC compiler as well or something different?

mjansen4857: It doesn't fail for me building for the RIO, as it does when running CI, so that would make sense. Although it should be building desktop for you as well unless you're explicitly telling it not to.

MrRedness: That would make sense. I didn't get to it last night, but I'll attempt to write a fix that works for Visual Studio as well this evening.

MrRedness: I have visual studio installed now, and I get the same issue. However, it remains even when testing code from an older commit, unless I missed something. I'll do more testing later in the weekend

mjansen4857: Does the main branch build? the main branch builds fine for me.

MrRedness: Yes, the main branch builds fine. The build was still failing despite me reverting the lines highlighted, because a similar issue arose on line 268. I fixed both by replacing emplace_back with push_back. It seems MSVC and gcc have different implementations for dealing with emplace_back and push_back (gcc simply calls emplace_back, but MSVC seems to handle this differently). Also, the wpi::json class is written in a very interesting way, because they handle types themselves as an enum `value_t` instead of as a template_argument. I'm sure this has its benefits, but it makes determining types a nightmare for the compiler. I'm pleasantly surprised that gcc can pull it off, but MSVC seems to be a little lacking, and gives up when it comes to using emplace_back for some reason. Anyways, for now, this builds.

MrRedness: It's also possible MSVC handles the construction of std::string differently, since it's complaining about not being able to determine whether to use const& or && (copy or move).

MrRedness: If you force it to use const & like this, it compiles just fine

`				
for (wpi::json::const_reference name : marker.at("names")) {
std::string const &ref = name;
names.emplace_back(name);
}
`

MrRedness: But for now I think just using what I committed should work best.

mjansen4857:
Alright. I updated the CI workflows to actually build for every platform instead of just running tests on linux. Once you update the branch and those run I'll merge.

MrRedness:
One thing I noticed today but forgot to implement, is that some of the includes are probably unnecessary/unused, such as <limits> in PPTraj.cpp and wpi::smallstring in PP.cpp

MrRedness:
That's probably something you'd be more familiar with anyway

mjansen4857:
Yeah I'm sure there are. There's probably some gradle plugin that can remove unused includes I just need to find it. The formatting plugin will do it for java but not C++.

PR #157, opened 2022-11-22
Added project specific settings

hadley31:
Adds project specific settings

Creates a file `.pathplanner/settings.json` in the project directory with the following contents:
```json
{
  "robotSize": {
    "width": 0.75,
    "length": 1.0
  },
  "holonomicMode": false,
  "generateJSON": false,
  "generateCSV": false,
  "pplibClient": true
}
```

### DIFF
--- a/lib/pages/home_page.dart
+++ b/lib/pages/home_page.dart
@@ -618,7 +618,7 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
         widget.prefs.setBool('generateJSON', json['generateJSON'] ?? false);
         widget.prefs.setBool('generateCSV', json['generateCSV'] ?? false);
       } catch (error) {
-        Log.error('An error occurred while loading settings');
+        Log.error('An error occurred while loading project settings', error);
       }
     }
 
--- a/lib/pages/home_page.dart
+++ b/lib/pages/home_page.dart
@@ -645,9 +645,9 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
     };
 
     settingsFile.writeAsString(encoder.convert(settings)).then((_) {
-      Log.debug('Wrote json to file');
+      Log.debug('Wrote project settings to file');
     }).catchError((err) {
-      Log.error(err);
+      Log.error('Error writing project settings', err);
     });
   }
 
--- a/lib/pages/home_page.dart
+++ b/lib/pages/home_page.dart
@@ -547,30 +547,6 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
     return buildFile.existsSync();
   }
 
-  void _loadSettingsFromPrefs() {
-    setState(() {
-      _robotSize = Size(
-        widget.prefs.getDouble('robotWidth') ?? 0.75,
-        widget.prefs.getDouble('robotLength') ?? 1.0,
-      );
-      _holonomicMode = widget.prefs.getBool('holonomicMode') ?? false;
-      _generateJSON = widget.prefs.getBool('generateJSON') ?? false;
-      _generateCSV = widget.prefs.getBool('generateCSV') ?? false;
-      _pplibClient = widget.prefs.getBool('pplibClient') ?? false;
-
-      if (_pplibClient) {
-        PPLibClient.initialize(widget.prefs);
-      } else {
-        PPLibClient.stopServer();
-      }
-    });
-  }
-
-  void _onProjectSettingsChanged() {
-    _loadSettingsFromPrefs();
-    _saveProjectSettingsToFile(_projectDir!);
-  }
-
   void _savePath(RobotPath path) async {
     if (_projectDir != null) {
       bool result = await path.savePath(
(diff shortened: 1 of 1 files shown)


mjansen4857:
I think this does make sense, and will do some testing with it.

One thing I would change is I believe it makes more sense to leave the pplibClient setting out of this. It might not work on some computers for a variety of reasons and would make sense to leave it disabled for them but still have it enabled on a main dev machine for example. Its also not really a robot specific setting like the size, holonomic, and generation settings are.

hadley31:
Makes sense! I will change that. What are your thoughts about the field image or the team color? Should I add those?

mjansen4857:
I think those are fine as is. "Team color" is really just the app theme color so it can be a per-person preference. Field image could make sense, but this increases complexity since you would need to support custom field images by storing them in the .pathplanner folder as well. And do checks whenever a project is open to copy over everyone added field images if not already. It probably makes sense to leave this out when like 99% of people are just going to use the default field image anyways.

mjansen4857:
Just so you know, since the formatting check failed, you can just comment `/format` and a bot will format everything and commit changes. You could also just format the specific file if you'd prefer.

lib/pages/home_page.dart
```diff
@@ -610,10 +578,83 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
...
+      }
+    });
+  }
+
+  void _loadProjectSettingsFromFile(Directory projectDir) {
+    File settingsFile = File(join(projectDir.path, _settingsDir));
+
+    if (settingsFile.existsSync()) {
+      try {
+        final fileContents = settingsFile.readAsStringSync();
```
mjansen4857, line 610: Since nothing from the settings file is actually needed immediately, you could make this method async, then use await to replace the sync methods with the async versions. This would stop this method from holding up initState and freezing the app while loading the file.

```suggestion
  void _loadProjectSettingsFromFile(Directory projectDir) async {
    File settingsFile = File(join(projectDir.path, _settingsDir));

    if (await settingsFile.exists()) {
      try {
        final fileContents = await settingsFile.readAsString();
```

lib/pages/home_page.dart
```diff
@@ -610,10 +578,83 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
...
+      settingsFile.createSync(recursive: true);
+    }
+
+    const JsonEncoder encoder = JsonEncoder.withIndent('  ');
+
+    Map<String, dynamic> settings = {
+      'robotSize': {
+        'width': _robotSize.width,
+        'length': _robotSize.height,
+      },
```
mjansen4857, line 641: I think you could just store these as `robotWidth` and `robotLength` to match the keys in shared preferences. It would also makes these two lines above cleaner:
```dart
widget.prefs.setDouble('robotWidth', json['robotSize']?['width'] ?? 0.75);
widget.prefs.setDouble('robotLength', json['robotSize']?['length'] ?? 1.0);
```

lib/pages/home_page.dart
```diff
@@ -610,10 +578,83 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
...
+
+        widget.prefs
+            .setDouble('robotWidth', json['robotSize']?['width'] ?? 0.75);
+        widget.prefs
+            .setDouble('robotLength', json['robotSize']?['length'] ?? 1.0);
+        widget.prefs.setBool('holonomicMode', json['holonomicMode'] ?? false);
+        widget.prefs.setBool('generateJSON', json['generateJSON'] ?? false);
+        widget.prefs.setBool('generateCSV', json['generateCSV'] ?? false);
+      } catch (error) {
+        Log.error('An error occurred while loading settings');
```
mjansen4857, line 621: Pass the caught error to the logger so it gets included in the log file. Also, it would be better to specify that the error was with project-spcific settings instead of settings in general.

```suggestion
        Log.error('An error occurred while loading project settings', error);
```

lib/pages/home_page.dart
```diff
@@ -610,10 +578,83 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
...
+      },
+      'holonomicMode': _holonomicMode,
+      'generateJSON': _generateJSON,
+      'generateCSV': _generateCSV,
+    };
+
+    settingsFile.writeAsString(encoder.convert(settings)).then((_) {
+      Log.debug('Wrote json to file');
+    }).catchError((err) {
+      Log.error(err);
```
mjansen4857, line 650: For the debug log, could you make the message a bit more descriptive so its easy to know where it came from when seeing it in the console. Also, you should add a message to the error log so that the error object is the second argument and gets formatted correctly.

```suggestion
      Log.debug('Wrote project settings to file');
    }).catchError((err) {
      Log.error('Error writing project settings', err);
```

mjansen4857 (changes requested):

lib/pages/home_page.dart
hadley31: Gotcha! I was in between, but I agree they shouldn't be needed right away, so async should be fine here

mjansen4857:
Other than the last change requested above, everything looks and works good.

lib/pages/home_page.dart
```diff
@@ -610,10 +578,79 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
...
+  void _loadProjectSettingsFromFile(Directory projectDir) async {
+    File settingsFile = File(join(projectDir.path, _settingsDir));
+
+    if (await settingsFile.exists()) {
+      try {
+        final fileContents = await settingsFile.readAsString();
+        final json = jsonDecode(fileContents);
+
+        widget.prefs.setDouble('robotWidth', json['robotWidth'] ?? 0.75);
+        widget.prefs.setDouble('robotLength', json['robotLength'] ?? 1.0);
```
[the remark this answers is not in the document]
hadley31: Ohh interesting, good catch!

PR #178, opened 2022-12-14
Use list of points instead of vararg for specifying on the fly trajectories.

brennenputh:
This PR changes the generatePath method to accept a list of PathPoints instead of taking two arguments and a vararg of PathPoints.

The reasoning behind this change is situations like my own, where I would like to be able to pass several points into a function to get a command that runs a path.  Using the PathPlanner generatePath method everywhere would mean unnecessary code duplication of the constraints (I'm only running one robot, the constraints don't change during runtime).  This causes me to have to write ugly code in order to accept a list of points, getting the first two by index and then calling subList and converting that sub list to an array.

The secondary reasoning is that the paths are simply reconstructed into an ArrayList at the end of the day already.  It doesn't make sense to convert away from a List back into a List.

As other options on this PR, I'm unsure whether it would be better to keep a vararg overload.  If this is desired, I can add one.  Also, it may be easier to add overloads that take Lists, however this adds a lot of unnecessary code for the Lists to be turned into arrays for varargs to be turned right back into a List.

Feel free to deny if this is actively not desired.

### DIFF
--- a/pathplannerlib/src/main/native/include/pathplanner/lib/PathPlanner.h
+++ b/pathplannerlib/src/main/native/include/pathplanner/lib/PathPlanner.h
@@ -92,11 +92,13 @@ public:
 	 * of what works well and what doesn't before you use this method in competition. Points positioned in weird
 	 * configurations such as being too close together can lead to really janky paths.
 	 *
-	 * @param constraints The max velocity and max acceleration of the path
+	 * @param maxVel The max velocity of the path
+	 * @param maxAccel The max acceleration of the path
 	 * @param reversed Should the robot follow this path reversed
 	 * @param points Points in the path
 	 * @return The generated path
 	 */
+	[[deprecated("Use generatePath(PathConstraints, bool, std::vector<PathPoint>) instead")]]
 	static PathPlannerTrajectory generatePath(
 			units::meters_per_second_t const maxVel,
 			units::meters_per_second_squared_t const maxAccel,
@@ -104,7 +106,6 @@ public:
 		return generatePath(PathConstraints { maxVel, maxAccel }, reversed,
 				points);
 	}
-	;
 
 	/**
 	 * @brief Generate a path on-the-fly from a list of points
@@ -128,8 +129,8 @@ public:
 	 * of what works well and what doesn't before you use this method in competition. Points positioned in weird
 	 * configurations such as being too close together can lead to really janky paths.
 	 *
-	 * @param constraints The max velocity and max acceleration of the path
-	 * @param reversed Should the robot follow this path reversed
+	 * @param maxVel The max velocity of the path
+	 * @param maxAccel The max acceleration of the path
 	 * @param points Points in the path
 	 * @return The generated path
 	 */
--- a/pathplannerlib/src/main/native/include/pathplanner/lib/PathPlanner.h
+++ b/pathplannerlib/src/main/native/include/pathplanner/lib/PathPlanner.h
@@ -133,7 +133,7 @@ public:
 	 * @param points Points in the path
 	 * @return The generated path
 	 */
-	 [[deprecated("Use generatePath(PathConstraints, std::vector<PathPoint>) instead")]]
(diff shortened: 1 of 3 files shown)


mjansen4857:
Instead of replacing the methods, I would prefer overloads. Using varargs makes it a bit cleaner to use in some cases so I'd like to keep that option.

For example:
```Java
generatePath(constraints, currentRobotPoint, loadingStationPoint);
```

vs

```Java
generatePath(constraints, List.of(currentRobotPoint, loadingStationPoint));
```

It just removes the need to create a list yourself. All the vararg overload would need to do is create the list for you.

brennenputh:
Added in latest commit.  I do have one question, what is the point of having a overload with the max velocity and max acceleration broken out instead of passed in as PathConstraints?  It seems like it adds a lot of overloads for not much gain.

brennenputh:
One other thing, would it be wise to implement the same change in C++?  I am inexperienced in C++ so I am unsure if there is a standard.

mjansen4857:
C++ could benefit from doing the same thing, just vectors and initializer lists instead of lists and varargs. I could take care of that though if you'd like.

brennenputh:
I'll get the C++ changes done, they don't look hard from a glance.  I'll also make sure to mark the separate vel/accel as deprecated.

brennenputh:
Looked around, I'm not sure what to use as the equivalent of deprecated in C++.  Other than that though, I believe it's done.

mjansen4857:
It looks like you can do deprecation in C++ like this:
```C++
[[deprecated("Use someOtherMethod() instead.")]]
void deprecatedMethod() {}
```

Haven't bothered to try to compile it with the roborio toolchain so if it fails to build then no big deal.

brennenputh:
Done.

pathplannerlib/src/main/native/include/pathplanner/lib/PathPlanner.h
```diff
@@ -72,6 +72,75 @@ class PathPlanner {
...
+	 * As you can't see the path in the GUI when using this method, make sure you have a good idea
+	 * of what works well and what doesn't before you use this method in competition. Points positioned in weird
+	 * configurations such as being too close together can lead to really janky paths.
+	 *
+	 * @param constraints The max velocity and max acceleration of the path
+	 * @param reversed Should the robot follow this path reversed
+	 * @param points Points in the path
+	 * @return The generated path
+	 */
+	static PathPlannerTrajectory generatePath(
```
mjansen4857, line 100: Missed a deprecation here

pathplannerlib/src/main/native/include/pathplanner/lib/PathPlanner.h
```diff
@@ -72,6 +72,75 @@ class PathPlanner {
...
+			units::meters_per_second_t const maxVel,
+			units::meters_per_second_squared_t const maxAccel,
+			bool const reversed, std::vector<PathPoint> const points) {
+		return generatePath(PathConstraints { maxVel, maxAccel }, reversed,
+				points);
+	}
+	;
```
mjansen4857, line 107: This looks like a random semicolon

mjansen4857 (changes requested):

auscompgeek:
Could we get another release with this change in? Currently looking at updating the RobotPy bindings for 2023 and noticed the `generatePath` method is unusable from our end.

virtuald:
To explain (and keep in mind for the future): RobotPy cannot use std::initializer_list since the list has to be known at compile time, so anything that uses that needs a different overload that takes a span or a vector or whatever otherwise RobotPy cannot use it. Thanks!

mjansen4857:
Sure thing. I will look through everything to make sure there are initializer list alternatives and then push a PPLib update at least.

mjansen4857:
OK update is now available. If there are any other issues with RobotPy support that need changes feel free to open an issue and I'll take a look.

PR #180, opened 2022-12-18
Replace HashMap with Map interface

### LINKED ISSUE #179: Implementations should be replaced with interfaces in parameter types.
It is generally not recommended to structure code to take in or return objects of an implementation type, rather the interface type. This lets users of the given api use whatever implementation they want. In SwerveAutoBuilder.java (and there may be more examples), the constructor takes in a HashMap<String, Command>, but Java convention is to take in Map<String, Command>.

hadley31:
Closes #179


hadley31:
Ah, had it right the first time 😄 Will it work if I edit?

mjansen4857:
I already linked it manually so it shouldn't make a difference. I'll take a look and merge tomorrow

PR #183, opened 2022-12-20
Updated pplib to use new Commands API

### LINKED ISSUE #176: CommandGroupBase class appears to be on the way out.
I was just looking at WPILib and noticed that a recent PR modified a bunch of command group stuff.  
[WPILib Pull Request 4728](https://github.com/wpilibsuite/allwpilib/pull/4728)

I don't think the changes will break PathPlanner in 2023, but I think the line below might need to be updated before CommandGroupBase disappears all together.  As far as I can tell it appears to be depreciated. I only found this line from CTRL+F "CommandGroupBase"

https://github.com/mjansen4857/pathplanner/blob/b57e0083ba9e31baee8a3aa814809177df995c6d/pathplannerlib/src/main/java/com/pathplanner/lib/auto/BaseAutoBuilder.java#L151

### LINKED ISSUE #181: StopEvent with empty event list
https://github.com/mjansen4857/pathplanner/blob/b57e0083ba9e31baee8a3aa814809177df995c6d/pathplannerlib/src/main/java/com/pathplanner/lib/auto/BaseAutoBuilder.java#L159

Seems that in a stop point with no events (just wait time), you would get an exception of array index out of bounds.

hadley31:
Closes #176 
Closes #181 

## Notable Changes
* If stopEvent.names is empty, just return a wait command
* Removed reliance on `CommandGroupBase`, opting for the newer `Commands` API
* Created a utility `CommandList` class. It is just a subclass of `ArrayList<CommandBase>` with some helpful methods to make converting between collections of Commands to arrays of commands a lot prettier and easier. Let me know your thoughts on this, but to me it seems like the most elegant solution
* Heavy reliance on the java stream api

Let me know if you like these changes, and I can work on the C++ implementation.

### DIFF
--- a/pathplannerlib/src/main/native/cpp/pathplanner/lib/auto/BaseAutoBuilder.cpp
+++ b/pathplannerlib/src/main/native/cpp/pathplanner/lib/auto/BaseAutoBuilder.cpp
@@ -1,12 +1,7 @@
 #include "pathplanner/lib/auto/BaseAutoBuilder.h"
 #include "pathplanner/lib/commands/FollowPathWithEvents.h"
 
-#include <frc2/command/SequentialCommandGroup.h>
-#include <frc2/command/ParallelCommandGroup.h>
-#include <frc2/command/ParallelDeadlineGroup.h>
-#include <frc2/command/InstantCommand.h>
-#include <frc2/command/WaitCommand.h>
-#include <frc2/command/FunctionalCommand.h>
+#include <frc2/command/Commands.h>
 
 using namespace pathplanner;
 
@@ -19,13 +14,13 @@ BaseAutoBuilder::BaseAutoBuilder(std::function<frc::Pose2d()> pose,
 
 frc2::CommandPtr BaseAutoBuilder::followPathGroup(
 		std::vector<PathPlannerTrajectory> pathGroup) {
-	std::vector < std::unique_ptr < frc2::Command >> commands;
+	std::vector < frc2::CommandPtr > commands;
 
 	for (PathPlannerTrajectory path : pathGroup) {
-		commands.emplace_back(followPath(path).Unwrap());
+		commands.emplace_back(followPath(path));
 	}
 
-	return frc2::SequentialCommandGroup(std::move(commands)).ToPtr();
+	return frc2::cmd::Sequence(std::move(commands));
 }
 
 frc2::CommandPtr BaseAutoBuilder::followPathWithEvents(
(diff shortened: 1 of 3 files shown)


JCodeShelver:
Use of Streams API comes with pros and cons. It can (depending on the individual) be easier for other programmers to read (given this is open-source, that's important), but also can be slower than other techniques (like for-each over a list, or a normal for-loop from 1-4 for swerve modules).

hadley31:
That is true. For this case, I don't think there is any reason to be concerned about performance, since these methods should only be called once when the robot boots. If these methods were called more frequently, I would probably steer away from streams, and opt for a faster approach. At robot init, loading and parsing the json file is taking the bulk of the time anyway.

mjansen4857:
Have you tested that these changes still produce a working auto command on a simulator or something yet? Just wanna know how closely I should look at stuff.

hadley31:
Yeah, they were just in a temp main method I wrote. I can try looking into that as well

hadley31:
Just finished testing with the sim. AutoBuilder is still working as expected 🙂

mjansen4857:
So, after looking through it, I think it may be beneficial for readability/maintainability to just go with something other than streams for the CommandList stuff. This way the Java and C++ versions can be essentially identical with some form of for loop. I think its fine if it were just java, but in C++ I think something similar would look significantly different. So, it would be a bit of a pain to make changes to both versions if they use very different APIs.

hadley31:
Ahh dang, very good point. I'll change it to something more agnostic

hadley31:
Changed it to use ArrayLists and for loops. I started making the changes in C++, but I may be in over my head. I will push what I have so far soon

pathplannerlib/src/main/native/cpp/pathplanner/lib/auto/BaseAutoBuilder.cpp
```diff
@@ -77,66 +72,64 @@ frc2::CommandPtr BaseAutoBuilder::wrappedEventCommand(
...
 			== PathPlannerTrajectory::StopEvent::ExecutionBehavior::PARALLEL_DEADLINE) {
-		frc2::CommandPtr deadline = frc2::InstantCommand().ToPtr();
-		if (m_eventMap.find(stopEvent.names[0]) != m_eventMap.end()) {
-			deadline = wrappedEventCommand(m_eventMap.at(stopEvent.names[0]));
-		}
-		events = frc2::ParallelDeadlineGroup(std::move(deadline).Unwrap(),
-				std::move(eventCommands)).ToPtr();
+		frc2::CommandPtr *deadline = &eventCommands.front();
+		eventCommands.erase(eventCommands.begin());
+		return frc2::cmd::Deadline(std::move(*deadline),
```
hadley31, line 99: Please fix this 😄

mjansen4857:
```C++
frc2::CommandPtr deadline = frc2::cmd::None();
if (m_eventMap.find(stopEvent.names[0]) != m_eventMap.end()) {
	deadline = wrappedEventCommand(m_eventMap.at(stopEvent.names[0]));
}
return frc2::cmd::Deadline(std::move(deadline), std::move(eventCommands));
```

The deadline command is already left out of the `eventCommands` list so you don't need to do that again.

Also, the final else if block in stopEventGroup has frc2::Command as the type in the vector instead of frc2::CommadPtr, this will fail to build

hadley31: Ahh, you right! Thanks

PR #186, opened 2022-12-22
First event on stop event to show deadline label in UI

hadley31:


lib/widgets/path_editor/cards/stop_event_card.dart
```diff
@@ -290,6 +290,17 @@ class _StopEventCardState extends State<StopEventCard> {
     );
   }
 
+  String _getEventLabel(int index) {
+    if (widget.stopEvent!.executionBehavior ==
+        ExecutionBehavior.parallelDeadline) {
+      if (index == 0) return 'Deadline';
```
mjansen4857, line 296:
```suggestion
      if (index == 0) return 'Deadline Event';
```

I think it makes sense to change the label to "Deadline Event" to try and keep it clear that this still comes from the event map.

mjansen4857 (changes requested):

lib/widgets/path_editor/cards/stop_event_card.dart
hadley31: Agreed

PR #238, opened 2023-02-02
Fix drivetrain type initialization

gcjurgiel:
Add initialization of drivetrain type to `BaseAutoBuilder` that a lack of had previously created issues for differential drive paths starting at non zero rotations


mjansen4857:
Formatting bot isn't working for your branch for some reason so you'll have to do that on your end: `./gradlew spotlessApply`

Also, the build is failing because of the initialization order, you'll want to move the initialization of the drive train type before the initialization of useAllianceColor.

gcjurgiel:
Thanks!

mjansen4857:
Just realized I totally forgot to release this lol, my bad. I'll work on getting that pushed today.

PR #247, opened 2023-02-12
Add Focused Selection Option

### LINKED ISSUE #245: Add the ability to navigate waypoints like is available for Marker
Selecting the correct waypoint when you have multiple on top of each other is difficult at best. It would be nice to have the same "navigation" options that are available when editing markers. So you can use a right or left arrow to go to the next or previous waypoint.

### LINKED ISSUE #248: Delete key shortcut to remove waypoints and markers

hadley31:
Closes #245 
Closes #248 

### Description
Adds an option to hide all waypoints that aren't directly connected to the selected waypoint. I still think there is work to be done here, and would appreciate feedback from others to make this as nice a feature as possible. 

### With focused selection disabled:
image

### With focused selection enabled:
image

### New setting:
image

### DIFF
--- a/lib/pages/home_page.dart
+++ b/lib/pages/home_page.dart
@@ -47,6 +47,7 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
   RobotPath? _currentPath;
   Size _robotSize = const Size(0.75, 1.0);
   bool _holonomicMode = false;
+  bool _focusedSelection = false;
   bool _generateJSON = false;
   bool _generateCSV = false;
   bool _pplibClient = false;
@@ -493,6 +494,7 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
               robotSize: _robotSize,
               holonomicMode: _holonomicMode,
               showGeneratorSettings: _generateJSON || _generateCSV,
+              focusedSelection: _focusedSelection,
               savePath: (path) => _savePath(path),
               prefs: widget.prefs,
             ),
@@ -626,6 +628,7 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
         widget.prefs.getDouble('robotLength') ?? 1.0,
       );
       _holonomicMode = widget.prefs.getBool('holonomicMode') ?? false;
+      _focusedSelection = widget.prefs.getBool('focusedSelection') ?? false;
       _generateJSON = widget.prefs.getBool('generateJSON') ?? false;
       _generateCSV = widget.prefs.getBool('generateCSV') ?? false;
       _pplibClient = widget.prefs.getBool('pplibClient') ?? false;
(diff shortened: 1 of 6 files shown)


mjansen4857:
Yeah, I ran into this issue a couple weeks ago but had no idea for the best way to handle it. I think I like this solution. One suggestion I do have is to draw the path outside of the "focused selection" with some opacity so they will appear a bit faded/darker. This will give some contrast between the segment of the path you're actually looking at and the rest of the path.

This doesn't really help with actually selecting a waypoint if you have a few near each other. I have ideas for next year to make the UI generally better for this stuff, but #245 should be a good temporary solution for that.

hadley31:
Cool, I will work on that tomorrow.

EddieDL:
Instead of having a persistent option would it be better to have a "hot key" to temporarily show the focused mode? Not sure if "while held" or "toggle" would be better, but I can see a ton of scenarios where you might want to switch between them often. The idea of going into settings all the time doesn't seem great.

Not that it should be done with this item, but the same functionality for Edit Markers would be cool as well. Not sure if it really should be the exact same, but overlapping paths, waypoints and markers are tough.

mjansen4857:
So I believe that this focused selection thing will still show all waypoints when you deselect a waypoint. So, I’m not sure how often switching it off an on would actually be done. 

Also, shortcuts aren’t a visible way to control the app. Most people don’t read the wiki so they won’t know this feature exists. If there is a visible setting then it is much more likely to actually get utilized

WarrenReynolds:
Wish List.
Another nice feature here, would be to scale the size of the dots in the GUI on the ends of the spline handles and robot front as you zoom in and out with the scroll wheel to make it easier to grab them when they are close to other handles ends.
And if you wanted to really go to town with this , scales the thicknesses of the path lines and robot rectangle as you zoom in as well.
Also a text box to manually enter the length on the currently selected control handle. That's the last thing that you cant specify numerically.

hadley31:
@mjansen4857 here is what the path looks like now:
image

And for double line:
image

EddieDL:
I love this feature! My only question is... when can I get it :P

mjansen4857:
Haven't dived into the code yet... Does this disable click detection for the hidden waypoints so they aren't accidentally selected?

hadley31:
> Haven't dived into the code yet... Does this disable click detection for the hidden waypoints so they aren't accidentally selected?

Yes, it does. I discovered that very quickly that the rendering and click detection were separate haha

mjansen4857:
@hadley31  So I didn’t test it enough before releasing it. Found out that you’re unable to click and drag the end point regardless of the setting being enabled or not. Could you take a look at that? I’m unable to at the moment.

EddieDL:
Thanks @hadley31 and @mjansen4857 for this update It really makes the tool WAY more usable.

PR #256, opened 2023-02-17
Generated javadocs webpage

thenetworkgrinch:
I generated the javadocs webpage based off of the javadoc comments in pathplanner lib to help with documentation. There are alot of missing comments, but it is a start. You can host the javadocs using github pages too https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-github-pages-site


mjansen4857:
I think I want to handle this a bit differently. I can set up some github actions workflows that can generate these docs on every commit and publish them with github pages in a separate branch. That way it won't clutter up the main branch.

mjansen4857:
Yeah I want them too just have to find the time to set up the actions. I definitely don't want to have to generate these manually every time a change is made.

Issue #207, opened 2023-01-08
Red Alliance - Invert Path Y Value


hadley31:
Add alliance parameter to PPLib classes, telling the path follower to invert the path y value if the alliance is Red

hadley31:
Third times the charm 🙂

hadley31:
Or possibly have the UI export two different paths, either in the same file or separate

mjansen4857:
I ink I want to avoid doing this in the UI if possible since it is a lot more effort to do it in a nice way. And exporting 2 different paths may be annoying to work with. I think doing it in the load path methods is really simple and the way to go.

WarrenReynolds:
I think this needs a lot more thought.  This isn't just a case of added a negative sign.  You are going to have to pass in the field width and alliance colour to the generate path method because coordinates of points and spine handles are all going to turn into (FieldWidth-Value) instead of Value. 

What have WPILib done with the whole single mirror instead of double mirror field setup thing. Does any body know how the coordinates of the April Tags are returned from the WPI Call?   Does this call take an Alliance colour or is Tag ID 1 always reported at the same location regardless of your alliance colour?

mjansen4857:
I think it’s safe to assume that everyone using 2023 PPLib will be playing on the charged up field so there’s not really a need to take in field width and what not. Idk what WPILib will do with pathweaver but I haven’t seen any updates to it since last year. 

The WPILib apriltags stuff has the ability to take alliance color into account when returning tag positions. However, this was likely designed with a normal field mirror so it will probably be updated.

gcschmit:
AprilTags are unique on the field (no ID is repeated). As a result, AprilTags themselves aren't directly impacted by the reflected field layout.

hadley31:
I thought about that too, but it took me a while to realize that the AprilTagFieldLayout will work no matter what. The positions will always be correct since they are just _relative to_ different points. It is just the autos that are affected.

WarrenReynolds:
This all hinges around do you call 0,0 your alliance wall on the right side regardless of if you are Red or Blue?

[...]

PR #263, opened 2023-02-25
Refactor rotational transformation for red alliance

PatribotsProgramming:
state.poseMeters.getRotation().times(-1) will return -PI when PI is inputted, but when the robot is on the opposing alliance, the rotation needs to be 0 degrees if 180 is inputted, to preserve the holonomic rotation facing the grid.

The cpp equivalent for this still needs to be made.

Co-authored-by: Alexander Hamilton <a.hamilton72006@gmail.com>

### DIFF
--- a/pathplannerlib/src/main/java/com/pathplanner/lib/PathPlannerTrajectory.java
+++ b/pathplannerlib/src/main/java/com/pathplanner/lib/PathPlannerTrajectory.java
@@ -122,8 +122,8 @@ public class PathPlannerTrajectory extends Trajectory {
 
       Translation2d transformedTranslation =
           new Translation2d(state.poseMeters.getX(), FIELD_WIDTH_METERS - state.poseMeters.getY());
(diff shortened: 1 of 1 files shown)


PatribotsProgramming:
I should also mention that #207 should be mirrored on the X axis and not the Y, as seen in [this Cheif Delphi post](https://www.chiefdelphi.com/t/pathplanner-trajectory-mirrors-on-x-instead-of-y/423757), meaning that I should have included this change in the PR.

GalexY727:
I had to include the java docs haha

hadley31:
Hi @PatribotsProgramming, please read discussion #261

mjansen4857:
> rotation needs to be 0 degrees if 180 is inputted

This is not correct. 0 Degrees is facing away from your driver driver station wall.

Also, the alliance transformation stuff is correct, so I will be closing this. If your coordinate system is different from what the WPILib standard is then you will need to implement the transformations yourself, since PPLib will follow the wpilib standards.

PR #264, opened 2023-02-27
Refactor alliance transform for C++

gcjurgiel:
Thank you to @PatribotsProgramming for fixing this in Java (#263) so that I could copy it for the C++ fix.

### DIFF
--- a/pathplannerlib/src/main/native/cpp/pathplanner/lib/PathPlannerTrajectory.cpp
+++ b/pathplannerlib/src/main/native/cpp/pathplanner/lib/PathPlannerTrajectory.cpp
@@ -398,7 +398,7 @@ PathPlannerTrajectory::PathPlannerState PathPlannerTrajectory::transformStateFor
 
 		frc::Translation2d transformedTranslation(FIELD_WIDTH - state.pose.X(),
 				state.pose.Y());
-		frc::Rotation2d transformedHeading = state.pose.Rotation() 
+		frc::Rotation2d transformedHeading = state.pose.Rotation()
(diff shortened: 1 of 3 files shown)


GalexY727:
Of course! We're happy to help.

WarrenReynolds:
Whats going on here with this PR?  To me this looks like it will break everything. Has WPILib given up on 0,0 being your alliance wall on the right hand side regardless of your alliance colour?

X values in paths shouldn't be touched. If you are Blue and you want to drive down the field away from your alliance wall 1 meter your odometry X value will go up by 1 metre.  If you Red and you drive down the field away from your alliance wall 1 meter you odometry X value will go up by 1 meter.  The X values in a path should not be changes.  It was correct before this PR.

gcjurgiel:
No you are correct and I just misunderstood the field coordinate system. Since an alliance dependent origin is what WPILib officially uses, this change and the Java version made in #263 should be removed.

I misattributed this as an issue because it seems to me that the `Feild2d` Widget in shuffleboard does not properly account for alliance, placing the robot graphic in the wrong position. This limit has not been issue in previous years since the field was rotational symmetric, so would be placed in the right position but on the wrong side of the field.

Additionally our teams vision based odometry system is not alliance dependent, finding the same position with the blue alliance as the origin regardless of our teams alliance, as is the origin for the aprilTags in the field CAD files. For this reason our team will have to change its transform function, but PathPlanner should not. It seems WPILib may want to reconsider this field coordinate system if non-rotationally symmetric fields continue in the future.

PR #282, opened 2023-03-14
Publicize more C++ things

rbiliff:
This basically just implements Pull #270 into C++ as well as publicizing the extra constructor for `PathPlannerTrajectory`. As is it's borderline impossible to create a trajectory tweaked from another trajectory (such as for custom origin flipping).


mjansen4857:
There is an issue in PathPlanner.cpp, you changed the generatePath signature to take in a vector of event markers but never actually use it. You should pass that vector to the created trajectory on line 178

rbiliff:
Good catch, completely glossed over that change in the java version 👍

PR #295, opened 2023-03-30
Add equality checking for PathPoints

rshah713:
One great feature of WPILib's `Pose2d` is the ability to [equality-check](https://github.com/wpilibsuite/allwpilib/blob/aa34aacf6eff05477400bbcc46a2e9e4014e0059/wpimath/src/main/java/edu/wpi/first/math/geometry/Pose2d.java#L272-L279
) the objects.

I've ported the logic for PathPoints.


mjansen4857:
Format bot won't work for you branch for some reason. You'll need to run `./gradlew spotlessApply` and push those changes

Issue #289, opened 2023-03-21
Display total runtime while editing trajectory


Nsl106:
This would be helpful when optimizing a path to be faster so you don't need to switch between tabs every time

PatribotsProgramming:
Please please please please please this would be so useful!!!

PatribotsProgramming:
This also brings up the issue that you cannot CTRL+Z after clicking "play" to a path, so you have to full commit to the path if you want to see how long it takes when trying to optimize.

PatribotsProgramming:
I think I am close to resolving this!

PatribotsProgramming:
Well, I completed it, but am having some trouble with formatting :(

PR #300, opened 2023-04-06
Display total runtime while editing trajectory #289

PatribotsProgramming:
I am having some trouble preventing the formatting errors, please look through [this commit](https://github.com/PatribotsProgramming/pathplanner/commit/90a09ae69c9b1c52713eb2aca9dd134d92f59379) to see the changes in full.

### DIFF
--- a/lib/pages/home_page.dart
+++ b/lib/pages/home_page.dart
@@ -502,6 +502,7 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
               showGeneratorSettings: _generateJSON || _generateCSV,
               focusedSelection: _focusedSelection,
               savePath: (path) => _savePath(path),
+              generateTrajectory: (path) => _generateTrajectory(path),
               prefs: widget.prefs,
             ),
           ),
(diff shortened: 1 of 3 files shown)


GalexY727:
Another thing that I should mention is that I have never worked with dart before, and therefore may not have implemented this as smoothly as possible. I have seen variables with question marks after them such as Bool? and if that means that the variable is assumed to have a possibility to be excluded from a call to a method, the smoother way of implementing this would be to simply call savePath, but have boolean overrides for writing to JSON/CSV. If these overrides are included, **exclude** making JSON/CSV, since that requires a lot more processing power and isn't optimized to run every loop.

GalexY727:
Not at all sure if there are any connections to this PR and recent commits and whatnot, but there may be a bug in this PR that makes it so that the undo and redo history isn't as reliable as it was before the addition of the total runtime timer in the edit_state.
I remember it working when submitting this PR, but since this is my first time using Dart I may be forgetting about something significant needed to preserve data such as the undo history in keys. 
The framework is there, though!

If it would be easier, there can always be a separate function. Instead of calling savePath() and not writing to the file, it just calls a calculation method that returns the total runtime, which should reduce the overall load on the app when paths are edited.

Again, still not sure if there is a connection between these changes and the loss of undo history... It _shouldn't be affected, right?_

GalexY727:
[heres the change log](https://github.com/mjansen4857/pathplanner/wiki/PathPlannerLib-Changelog)

mjansen4857:
Just a heads up, this won't get merged since there are some reworks in progress that would make these changes invalid. 

I'll be bringing this feature in with the reworks though. I'll leave this and the issue open until that stuff gets merged.

GalexY727:
Thanks! I hope it aids in the future

mjansen4857:
Closing this as this functionality is included in #317

Issue #303, opened 2023-04-14
Map arrow keys to move path point


Chris2fourlaw:
Feature request: For a selected path point, allow the arrow keys to translate the point up-down-left-right

mjansen4857:
I'd like to keep keyboard traversal enabled. So, instead of using the arrow keys to translate a selected waypoint, this will be implemented as using the arrow keys to increment the value in a text field. This also has the benefit of working for things besides position. Such as heading, rotations, etc.

Issue #113, opened 2022-09-06
Tab key doesn't save values before moving the focus.


WarrenReynolds:
Not really an issue or bug, more of a feature request. 
I always use the TAB key to move to the next field after updating a value.  At the moment if you do this the value in a text box doesn't get updated unless you press the ENTER key.  I'm so used to the ENTER key being used to accept the current change and close the dialog that I never press it when updating values, so I find myself always making mistakes in this area.

I don't know what Flutter's default behaviour is around dialog boxes, but if an options exists to make the TAB key update the value before it moves the focus to the next field this would be a nice option to have in the settings for path planner.

This is only a suggestion to make a great program even better.

mjansen4857:
So I looked into this and there's currently a bug with flutter that prevents me from knowing that the tab key was pressed (in order to update the path value) while still maintaining the behavior of the tab key switching focus to the next text field.

I'll come back to this once that gets fixed.

isaec:
Alternatively, after a bit of debouncing changes could be written to file without waiting for the focus to switch?

Seems slightly counterintuitive to only write changes when a new element is focused.

mjansen4857:
This will now work for number text fields after #328. But, I did not do this for any name text fields since renaming files and such is something I want to have intent behind it to prevent you from renaming a file to something weird when you click off of a text field and not notice that the name changed.

PR #307, opened 2023-04-18
Move points with arrow keys

towner-10:
- Added CTRL-S to save path after moving with keys
- Added CallbackShortcuts for moving with arrow keys

This solves #303 

The bottom EditorMode UI will select some elements when using the arrow keys, but after 2 clicks, the arrow key movement will work. This needs to be looked into for alternative keys for the EditorMode shortcuts.

### DIFF
--- a/lib/widgets/path_editor/editors/edit_editor.dart
+++ b/lib/widgets/path_editor/editors/edit_editor.dart
@@ -87,13 +87,71 @@ class _EditEditorState extends State<EditEditor> {
               removeWaypoint(_selectedWaypoint!);
             }
           },
-          child: Stack(
-            key: _key,
-            children: [
-              _buildEditor(),
-              _buildWaypointCard(),
-              _buildGeneratorSettingsCard(),
-            ],
+          child: KeyBoardShortcuts(
+            keysToPress: {
+              (Platform.isMacOS)
+                  ? LogicalKeyboardKey.meta
+                  : LogicalKeyboardKey.control,
+              LogicalKeyboardKey.keyS
+            },
+            onKeysPressed: () {
+              widget.savePath(widget.path);
+            },
+            child: CallbackShortcuts(
+              // Use arrow keys to move selected waypoint
+              bindings: <ShortcutActivator, VoidCallback>{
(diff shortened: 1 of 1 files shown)


mjansen4857:
As a heads up, this won't be merged as-is since there are some reworks in progress that would make the changes invalid. I might get this included with those reworks but if not it would probably be pretty easy to adapt for the new structure. I'll keep this updated either way but this is still a ways off.

WarrenReynolds:
> * Added CallbackShortcuts for moving with arrow keys

Is CallbackShortcuts the key to capture the Tab keypress in the Tab not saving values issue. https://github.com/mjansen4857/pathplanner/issues/113

I've sort of got used to pressing enter, but it still gets me from time to time.

mjansen4857:
The issue isn't capturing the tab key, its applying the changes without any confirmation input from the user. It will take some changes to get the app to actually handle this without being super annoying.

mjansen4857:
So after thinking about this a bit I don't think this quite the route I want to go down. Doing so would mean disabling keyboard navigation (I don't believe flutter allows disabling it just for the arrow keys and not the tab key) which has become a pretty nice feature to have in the new reworks. You can now completely edit a path without ever touching the mouse if you want.

I think a better solution to sort of accomplish this would be to use the up/down arrow keys while focused on a text field to increment its value. So, I'll close this PR in favor of that solution.

WarrenReynolds:
I sort of feel when adjusting angle field values with the arrow keys it should go to the nearest degree on the first press and then adjust by whole degrees on subsequent presses. This makes it easy to fine adjust to specific positions like 45, 90, 0, etc. Sort of like what Adobe illustrator does in its positional text boxes.  It really probably doesn't matter, as what is 0.5 of degree between friends, also happy for it to stay as is.  

BTW: I love the fact that the tab key now works.  I'm a bit old school, so I love navigating around on keyboards where I can.

mjansen4857:
Yeah that makes sense. I wanted to expand on it a bit with holding modifier keys to increment by different values anyways so I can make that change when I get around to that.

PR #310, opened 2023-06-09
Include generated Javadoc on GitHub Pages for PathPlannerLib

jasonli0616:
An idea for making development with PathPlannerLib easier.

The [wiki page for PathPlannerLib usage](https://github.com/mjansen4857/pathplanner/wiki/PathPlannerLib:-Java-Usage), though useful, is insufficient for teams looking to make the most out of the library.

The Javadoc is available through the [compiled version of PathPlannerLib](https://github.com/3015RangerRobotics/3015RangerRobotics.github.io/tree/main/pathplannerlib/repo/com/pathplanner/lib/PathplannerLib-java), but is not readily available on any webpage.

This can be added to the `docs` directory on this repo, and hosted for free through [GitHub Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-github-pages-site). A [GitHub Actions workflow](https://docs.github.com/en/actions/using-workflows) can be set up to automatically update this Javadoc webpage for a new release.

As a demo, I have forked this repo, and hosted the PathPlannerLib Javadoc page: https://jasonli0616.github.io/pathplanner/

PR #313, opened 2023-07-16
Create a GitHub Actions workflow to publish generated Javadoc to GitHub Pages

jasonli0616:
As pitched in https://github.com/mjansen4857/pathplanner/pull/310

# Overview

The GitHub Actions workflow is triggered based on a release (does not include drafts, or pre-release). Once a release is published (either directly, from a draft, or from a pre-release), it will assume that it's the latest version of PathPlannerLib.

It will build PathPlannerLib, and push the Javadoc directory to the `gh-pages` branch. It is not necessary to create a new branch for each version, because GitHub Pages can only host one version, and the generated Javadocs do not have versions built in. It will always be the latest version.

## Working demo

You can see a demo in my fork on the repo, at https://github.com/jasonli0616/pathplanner and https://jasonli0616.github.io/pathplanner/docs/java/

# How to implement

Because the workflow must push to the repo, it requires a bit of configuring on your (repo owner's) end to get it working.

## 1. [Deploy keys](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys#deploy-keys)

The workflow uses the [s0/git-publish-subdir-action](https://github.com/s0/git-publish-subdir-action) dependency to push a directory (in this case, the `pathplannerlib/build/docs/javadoc` directory) to a branch on a repo.

You can follow [their instructions](https://github.com/s0/git-publish-subdir-action#usage-with-deploy-keys) to set up the deploy keys.

In short:

1. Run `ssh-keygen -t ed25519 -C "your_email@example.com"`, substituting your email **(leave passphrase empty, when prompted)**
2. Add a secret in `(repo) > Settings > Secrets and variables > Actions > New repository secret`, name it `DEPLOY_PRIVATE_KEY` and paste in the private key
3. Add a deploy key in `(repo) > Settings > Deploy keys > Add deploy key` and paste in your public key

## 2. Configure GitHub Pages

The GitHub Pages must be configured to be hosted from the `gh-pages` branch, at the `/(root)` directory.

**See image for reference:**

Screenshot of GitHub Pages configuration

# Conclusion

The GitHub Actions workflow should now be set up to automatically build and publish the Javadoc on GitHub Pages when a new stable release is published.

Let me know your thoughts and if there are any tweaks or changes that you would prefer.

### DIFF
--- a/.github/workflows/publish-javadoc.yml
+++ b/.github/workflows/publish-javadoc.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Deploy generated Javadoc to GitHub Pages
         uses: s0/git-publish-subdir-action@v2.6.0
         env:
-          REPO: git@github.com:jasonli0616/pathplanner.git #change to git@github.com:mjansen4857/pathplanner.git before merge
+          REPO: git@github.com:mjansen4857/pathplanner.git
           FOLDER: pathplannerlib/build/docs/javadoc
           BRANCH: gh-pages
           TARGET_DIR: docs/java
--- a/.github/workflows/publish-javadoc.yml
+++ b/.github/workflows/publish-javadoc.yml
@@ -23,3 +23,4 @@ jobs:
           TARGET_DIR: docs/java
           SSH_PRIVATE_KEY: ${{ secrets.DEPLOY_PRIVATE_KEY }}
           MESSAGE: Publish Java docs from PathPlannerLib
+          SQUASH_HISTORY: true
--- a/.github/workflows/publish-javadoc.yml
+++ b/.github/workflows/publish-javadoc.yml
@@ -1,8 +1,7 @@
 name: Publish PathPlannerLib Javadoc
 
 on:
-  release:
-    types: [released]
+  workflow_dispatch
 
 jobs:
   build-and-deploy:
--- a/.github/workflows/publish-javadoc.yml
+++ b/.github/workflows/publish-javadoc.yml
@@ -8,6 +8,8 @@ jobs:
   build:
     runs-on: ubuntu-22.04
     steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
       - name: Build PathPlannerLib
         working-directory: pathplannerlib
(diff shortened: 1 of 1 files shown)


mjansen4857:
Could you have the docs be put in a `docs/java` subdirectory? This way its easily expandable C++ and Python docs in the future. The URL to browse the docs would then become https://jasonli0616.github.io/pathplanner/docs/java as well which is more descriptive.

jasonli0616:
Yep, the [change has been made](https://github.com/jasonli0616/pathplanner/commit/6d3da4e635454d9641f3e72784a1257d43e9bad7), it's [building](https://github.com/jasonli0616/pathplanner/actions/runs/5570096567) and should be updated in a couple minutes

mjansen4857:
Also, it should probably be triggered manually with `workflow_dispatch` instead of a new release since the releases on this repo are just for the GUI, so they don't correspond to PPLib releases.

.github/workflows/publish-javadoc.yml
```diff
@@ -0,0 +1,25 @@
...
+            ./gradlew publish -PreleaseMode
+      - name: Deploy generated Javadoc to GitHub Pages
+        uses: s0/git-publish-subdir-action@v2.6.0
+        env:
+          REPO: git@github.com:jasonli0616/pathplanner.git #change to git@github.com:mjansen4857/pathplanner.git before merge
+          FOLDER: pathplannerlib/build/docs/javadoc
+          BRANCH: gh-pages
+          TARGET_DIR: docs/java
+          SSH_PRIVATE_KEY: ${{ secrets.DEPLOY_PRIVATE_KEY }}
+          MESSAGE: Publish docs from PathPlannerLib version ${{ github.event.release.tag_name }}
```
mjansen4857, line 25: Last thing I can think of is this version label probably won't work anymore since its a manual dispatch, and it would be the GUI version anyway. A general "Publish PathPlannerLib java docs" commit will be fine I think.

Other than that, looks good to me.

PR #314, opened 2023-07-17
Link Javadoc to external sources (Java API, WPILib API)

jasonli0616:
A continuation of https://github.com/mjansen4857/pathplanner/pull/313, one last thing:

This should now link the Java docs and WPILib docs in the PathPlanner docs, where it's referenced.

### DIFF
--- a/pathplannerlib/build.gradle
+++ b/pathplannerlib/build.gradle
@@ -150,6 +150,12 @@ spotless {
     }
 }
 
+javadoc {
+    options {
+        links 'https://docs.oracle.com/en/java/javase/11/docs/api/', 'https://github.wpilib.org/allwpilib/docs/release/java/'
+    }
+}
+
 apply from: 'publish.gradle'
 
 wrapper {


mjansen4857:
👍 Just as a heads up if you're planning on doing any other PPLib/docs related stuff... A lot will be changing before the first 2024 release so you're probably better off waiting so it doesn't just get overwritten soon.

jasonli0616:
Got it

The stuff from today (publishing Javadoc) should be fine?

mjansen4857:
Yup. Just the code/actual javadocs will be reworked.

WarrenReynolds:
> 👍 Just as a heads up if you're planning on doing any other PPLib/docs related stuff... A lot will be changing before the first 2024 release so you're probably better off waiting so it doesn't just get overwritten soon.

So what's in the pipe line?  Do you have a roadmap of the new plans for PP? Please tell.

mjansen4857:
Not a lot of details yet since I'm still working through a lot of it but the main big things are:
* New pathfollowing/generation system that works on the fly instead of pre-generating paths. I think this is better than the old version for a few reasons but it was mainly done to better support the next thing:
* Automatic pathfinding/obstacle avoidance in PPLib
* Modular paths and "autos" that will just be multiple paths chained together with commands in between. This is replacing the concept of "stop points". Now if you make multiple autos that re-use the same path, updating that path will also update all of the autos that use it.

PR #317, opened 2023-08-07
2024 Reworks

### LINKED ISSUE #126: C++ PathPlannerLib Server
The socket headers that would be used for the PathPlannerLib server are linux specific, so as far as I'm aware it cannot be implemented easily without removing simulation support for Windows and macOS, or using a cross-platform socket library which I would like to avoid. If anyone is able to match the functionality of the Java server for all platforms, feel free to contribute. 

The PathPlannerLib server functionality should be converted to use the NetworkTables 4 REST API when/if that is released.

### LINKED ISSUE #138: Markers list
create a list of markers (per project, not per path), then in the markers editor you can choose a marker from a dropdown or just type a new name.

The reason is to prevent typos/inconsistencies in markers name, and reduce the chances of incompatibility with robot code.
(E.g., having multiple usages of “Open Intake” marker)

### LINKED ISSUE #163: At Stop Point Allow For Un-smooth Curvature
The straight curvature line from the PathPlanner UI

Would it make sense for at a stop point to allow for the curvature lines to not be one and to allow for it to be separate on each side of it?

### LINKED ISSUE #210: PathPlanner Immediately quits itself on M1 Mac
Downloaded from App Store.
Launch, see a window for a micro second then the app quits.
I don't think it crashes because I don't get a crash window.

I had an old version installed from 2020, it is possible there is a cache folder or something I'd need to delete to get the new version running?

### LINKED ISSUE #226: Window overflows on Linux HIDPI configuration
Hello,

On linux, the window overflows on startup.

Error:
```
flutter: ┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
flutter: │ A RenderFlex overflowed by 159 pixels on the bottom.
flutter: ├┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄
flutter: │ #0   Log.log (package:pathplanner/services/log.dart:40:15)
flutter: │ #1   Log.error (package:pathplanner/services/log.dart:63:5)
flutter: │ #2   main.<anonymous closure>.<anonymous closure> (package:pathplanner/main.dart:24:13)
flutter: │ #3   FlutterError.reportError (package:flutter/src/foundation/assertions.dart:1184:14)
flutter: │ #4   DebugOverflowIndicatorMixin._reportOverflow (package:flutter/src/rendering/debug_overflow_indicator.dart:239:18)
flutter: ├┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄
flutter: │ ⛔ Flutter Error
flutter: └───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

======== Exception caught by rendering library =====================================================
The following assertion was thrown during layout:
A RenderFlex overflowed by 159 pixels on the bottom.

The relevant error-causing widget was: 
  Column Column:file:///home/jackr/Projects/StudioProjects/pathplanner/lib/pages/welcome_page.dart:72:23
The overflowing RenderFlex has an orientation of Axis.vertical.
The edge of the RenderFlex that is overflowing has been marked in the rendering with a yellow and black striped pattern. This is usually caused by the contents being too big for the RenderFlex.

Consider applying a flex factor (e.g. using an Expanded widget) to force the children of the RenderFlex to fit within the available space instead of being sized to their natural size.
This is considered an error condition because it indicates that there is content that cannot be seen. If the content is legitimately bigger than the available space, consider clipping it with a ClipRect widget before putting it in the flex, or using a scrollable container rather than a Flex, like a ListView.

The specific RenderFlex in question is: RenderFlex#ddcde relayoutBoundary=up7 OVERFLOWING
...  parentData: offset=Offset(187.5, 0.0); flex=null; fit=null (can use size)
...  constraints: BoxConstraints(0.0<=w<=Infinity, 0.0<=h<=304.0)
...  size: Size(265.0, 304.0)
...  direction: vertical
...  mainAxisAlignment: center
...  mainAxisSize: max
...  crossAxisAlignment: center
...  verticalDirection: down
```
Screenshot of overflow: 
Screenshot_20230122_114037

This many be due to my Hi-DPI configuration (3840x2160).  If I set the `maximumSize` to my resolution, I can resize the window larger and then nothing overflows.  I believe there should be a way to remedy this, as hardcoded resolution sizes are problematic for 4K displays (`main.dart:27`):
```dart
WindowOptions windowOptions = WindowOptions(
        size: const Size(1280, 720),
        minimumSize: const Size(640, 360),
        maximumSize: const Size(3840,2160),  // this change allows me to resize to my display
        center: true,
        title: 'PathPlanner',
        titleBarStyle:
            Platform.isMacOS ? TitleBarStyle.normal : TitleBarStyle.hidden,
      );
```
This is the solution I am using PathPlanner with.  Theoretically this should adapt to all screen sizes, but a report from the API that isn't 16:9 (which is possible) could allow the screen to resize larger than it is allowed to.  I really don't know the errors that could happen:
```dart
        maximumSize: WidgetsBinding.instance.window.physicalSize,
```

This is a great app, I wish I could open this issue with a solution that is a lot less hacky than this.

### LINKED ISSUE #260: Newly created paths are appended to sidebar instead of being alphabetically sorted
1. Create a new path
2. Observe that it is at the bottom of the sidebar, regardless of its name
3. Close Pathplanner
4. Open Pathplanner
5. Observe that all paths, including the new one, are alphabetically sorted

### LINKED ISSUE #274: Add Angular Velocity/Acceleration Constraints
The feedforward values seem to be way too high which causes our robot to rotate much faster than we are comfortable with.

### LINKED ISSUE #285: Add option to run events and path in parallel
An option should be added to allow events and the path to run in parallel.

Consider the following case, assuming a robot with a long boom-like arm:

- A mid-path stop point with "closeGripper" and "pivotHome" events is created
- When the robot arrives at this stop point, it has its arm in a ground pickup position
- Due to mechanical specifics, attempting to pivot the arm without retracting the boom may cause damage from this position
- "pivotHome" is a sequential command group where the boom is retracted a certain amount before the arm pivots

As it currently stands, the robot waits for the "pivotHome" event to finish before continuing with the path. This takes many seconds. Having the option (a checkbox under "Execution Behavior"?) to tell PathPlanner to run the events and the path in parallel would be great.

For a specific implementation that might be ideal:
- wrapper ParallelCommandGroup
  - PathPlanner followTrajectory
  - User event commands

### LINKED ISSUE #289: Display total runtime while editing trajectory

### LINKED ISSUE #301: Incomplete or corrupt paths cause PathPlanner to crash
Because PathPlanner auto opens in a workspace, if it tries to load a corrupt or broken path, it causes the app to crash until the path file is manually deleted instantly. There should be some checks in place that disable a path if it doesn't load instead of hitting the entire app as it means that progress on path development can be easily lost (as in my case where I had to remake the path from scratch).

### LINKED ISSUE #304: Allow hiding of paths/points
Feature request:
Allow points/multi-paths to be hidden

Scenario:
When developing a large auto, showing all of the paths can get confusing. Being able to hide certain points or paths would be helpful

Implementation idea:
For autos with stop points, split each path in the path group into a "layer". I'm thinking like Adobe Illustrator, where each layer can be hidden. Color coding the path and points for each layer would also be useful

### LINKED ISSUE #311: return isReversed from getConstraintsFromPath
It seems that isReversed is getting set as a constraint for the path, but PathPlanner.getContraintsFromPath only returns maxVelocity and maxAcceleration. It would be useful to get that from GUI  via .path file and leverage automatically.

### LINKED ISSUE #316: Update needed for 2024 wpilib CommandBase change
wpilibsuite/allwpilib#5392 merged CommandBase into Command. A find and replace of CommandBase with Command in PathPlannerLib will be needed to compile with WPILib 2024.

mjansen4857:
Majority of the reworks for the 2024 release.

Fixes #316, closes #311, closes #304, fixes #301, closes #289, closes #285, closes #274, closes #260, fixes #226, fixes #210, closes #163, closes #138, closes #126

Supersedes #300 

TODO:
- [X] GUI reworks
- [X] Java PPLib reworks
- [X] C++ PPLib reworks (missing pathfinding for now)
- [X] Test Java PPLib in robot project
- [x] Documentation
- [x] Finish GUI unit tests
- [x] Test (and probably fix) C++ PPLib in robot project


jonahsnider:
Really exciting changes, looking forward to trying everything out this offseason! Thanks so much for helping create such a useful tool for FRC teams ❤️

PR #394, opened 2023-10-10
feat: add edit path button in auto editor

### LINKED ISSUE #393: Add button to edit paths from auto editor

hadley31:
Closes #393

### DIFF
--- a/lib/pages/auto_editor_page.dart
+++ b/lib/pages/auto_editor_page.dart
@@ -1,6 +1,5 @@
 import 'package:flutter/material.dart';
 import 'package:pathplanner/auto/pathplanner_auto.dart';
-import 'package:pathplanner/pages/project/project_page.dart';
 import 'package:pathplanner/path/pathplanner_path.dart';
 import 'package:pathplanner/services/pplib_telemetry.dart';
 import 'package:pathplanner/widgets/conditional_widget.dart';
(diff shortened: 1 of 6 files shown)


mjansen4857:
So one thing I noticed that might be a bit nicer is if instead of using the `OpenPathCallback` you made that passes a build context around, just pass the path name. You should be able to access the project page context within the _openPathByName method by using `this.context`;

hadley31:
> So one thing I noticed that might be a bit nicer is if instead of using the `OpenPathCallback` you made that passes a build context around, just pass the path name. You should be able to access the project page context within the _openPathByName method by using `this.context`.


Good call! I was hoping you'd suggest a better alternative 😄 I was under the impression I'd need at least the auto editor page context for it to work as I expected, but it makes sense that you could just use the project page context.

> Edit: Also, these changes are difficult to properly test since it depends on some coupling between the project page, auto editor, and path editor. I still haven't gotten around to doing a full integration test like that, so you can ignore those missed lines in project_page. Covering the icon button in path_command_widget shouldn't be too hard though. There should be plenty of examples in there if you're unfamiliar with flutter testing.


Okay! I'll write some tests for it tomorrow when I get some time.

hadley31:
TODO: If path is renamed in path editor pushed by the auto editor page, name change is not reflected unless user returns to projects page, and re-opens auto page. *DO NOT* reorder paths until this is fixed - app crashes and auto file gets wiped.

hadley31:
Going to close this for now. Haven't had the time to get around to it unfortunately, and from what I remember, there was a lot of Navigator context hoops to jump through to make this work without any looming bugs associated with renaming paths/commands. I'm thinking the slight productivity boost isn't worth the trouble.

PR #440, opened 2023-10-28
Add trim bar to preview seek bar

binex-dsk:
Currently broken because I have no clue how to use Flutter, but I got part of the way there so I'm making this PR now.

( The trim bar exists as RangeSlider, but the thumbs do not move. I have no clue why. :) )

### DIFF
--- a/lib/widgets/editor/preview_seekbar.dart
+++ b/lib/widgets/editor/preview_seekbar.dart
@@ -13,6 +13,8 @@ class PreviewSeekbar extends StatefulWidget {
 }
 
 class _PreviewSeekbarState extends State<PreviewSeekbar> {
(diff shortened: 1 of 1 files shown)


mjansen4857:
So I'm going to want this to be integrated into the seek bar itself, instead of adding another separate slider. Like how trimming video on a phone works would be a good example. This is why I said it would need a lot more custom work, since it would need to be a custom slider that integrates both the functionality of `RangeSlider` and `Slider` together.

binex-dsk:
Ah that makes more sense. Due to my lack of Dart experience this will be a no-go for me...

mjansen4857:
No problem, I'll open an issue for it and take it on at some point

PR #505, opened 2023-12-11
Initial Choreo Interoperability

mjansen4857:
* Run Chore trajectories in PPLib
* Create autos using Choreo trajectories in GUI


tbeadle:
I had never heard of Choreo before, so thanks for this. It looks pretty cool. I know you've been cranking out stuff like crazy, but since I'm very interested in this feature, are you planning on adding something to the wiki about it? Thanks again for all the amazing work!

PR #520, opened 2023-12-21
Fix a cyclical import in the Python library

EWall25:
Previously, a cyclical import between `trajectory.py` and `path.py` prevented the user from importing most of the library. This PR fixes the issue by changing the offending import to only occur while type-checking.

### DIFF
--- a/pathplannerlib-python/pathplannerlib/trajectory.py
+++ b/pathplannerlib-python/pathplannerlib/trajectory.py
@@ -5,10 +5,11 @@ from dataclasses import dataclass, field
 from wpimath.geometry import Translation2d, Rotation2d, Pose2d
 from wpimath.kinematics import ChassisSpeeds
 from wpimath import inputModulus
-import pathplannerlib.path as p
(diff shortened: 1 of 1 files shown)


mjansen4857:
Are you seeing actual errors for this? If so, what python version/platform? Importing path as a named module prevented the cyclical import for me and I'm able to run everything fine.

EWall25:
When I was testing last night, I got an ImportError upon importing `pathplannerlib.path` into my module. Running on Windows with Python 3.11.

I just tested again on the main branch, and somehow the error from last night is gone without me changing anything. You can probably close the PR without merging.

mjansen4857:
Huh. Were you testing before a full release was made? That problem was in there for a while before I actually started testing things.

EWall25:
Yeah, I'm not sure. I might have pulled from an outdated tag, realized it was "broken," pulled from the latest version, then tried to fix the issue without realizing it had already been fixed. Wierd.

PR #566, opened 2024-01-20
Fixed bug with Choreo path's not mirroring rotation's properly

Codehc:


mjansen4857:
The same issue is in the C++ version as well if you want to include that in this PR.

https://github.com/mjansen4857/pathplanner/blob/ed6efd284e5f23b6431fa1ffdcf501cb90e5058d/pathplannerlib/src/main/native/cpp/pathplanner/lib/path/PathPlannerPath.cpp#L730

Codehc:
Fixed that. I forgot cpp existed :p

PR #575, opened 2024-01-25
Initialize m_enabled to true in PPHolonomicDriveController

AJahueyM:
We are currently using PathPlanner for our robot, we are testing the code using a Gazebo simulation. The simulated robot didn't seem to turn at all, while the real robot did it without issue. We looked into it and found that the m_enabled check in PPHolonomicDriveController was failing, we believe it is because the variable is not assigned anything, leaving the compiler to decide its value. On Java the variable is set to true when its declared.

### DIFF
--- a/pathplannerlib/src/main/native/include/pathplanner/lib/controllers/PPHolonomicDriveController.h
+++ b/pathplannerlib/src/main/native/include/pathplanner/lib/controllers/PPHolonomicDriveController.h
@@ -105,7 +105,7 @@ private:
 	rpsPerMps_t m_mpsToRps;
 
 	frc::Translation2d m_translationError;
-	bool m_enabled;
+	bool m_enabled = true;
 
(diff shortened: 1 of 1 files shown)


mjansen4857:
Huh. Don't know why I never noticed that. Can you move the initialization to the constructor in the cpp file [here](https://github.com/mjansen4857/pathplanner/blob/07b821977e007136774fe1f6b295a0408557f9f4/pathplannerlib/src/main/native/cpp/pathplanner/lib/controllers/PPHolonomicDriveController.cpp#L16) instead of the h file? I like to keep it all in the same place.

mjansen4857:
Actually nvm, that's fine in the h file

mjansen4857:
When this gets merged, I'm just going to replace the current 2024.1.5 release with the new build since I just pushed 2024.1.5 today. So, if you are already updated to that version you'll have to delete it from your gradle cache.

AJahueyM:
Okay, thanks!

PR #578, opened 2024-01-26
Fix idempotency and flipping issue with PathfindingCommand

AJahueyM:
We are currently linking a Pathfind and Follow Path command to a button on our robot. We got it working great for the blue side of the field, but on the red side the Pathfinding would target the wrong angle. Once the actual follow path command started the heading got corrected properly. It seems like the heading issue was caused because m_goalEndState was not updated after m_targetPose was flipped. 

Regarding the idempotency issue, we noticed that using the command several times caused the heading to flip each time PathfindingCommand::Initialize was run, we just added a check so its only flipped once.

### DIFF
--- a/pathplannerlib-python/pathplannerlib/commands.py
+++ b/pathplannerlib-python/pathplannerlib/commands.py
@@ -275,6 +275,7 @@ class PathfindingCommand(Command):
     _timer: Timer = Timer()
     _targetPath: Union[PathPlannerPath, None]
     _targetPose: Pose2d
+    _originalTargetPose: Pose2d
     _goalEndState: GoalEndState
     _constraints: PathConstraints
     _poseSupplier: Callable[[], Pose2d]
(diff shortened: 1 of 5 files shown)


mjansen4857:
Doing the `initializedPath` boolean would break compatibility with hot reload and the ability to change alliance colors while testing. A better solution is to move the logic that initially creates the `m_targetPose` and `m_goalEndState` fields out of the constructor and into initialize, then do any flipping if needed. 

Or, it could save the target pose as `m_originalTargetPose` in the constructor. Then set `m_targetPose` and `m_goalEndState` based on either the original pose or the flipped pose in initialize. This is similar to how the normal FollowPathCommand handles flipping the path.

This would also need to be changed for Java and Python if you're able to.

AJahueyM:
Okay, will look into it

mjansen4857:
Actually I think the second option I told you is the only one that would work. Now that I'm looking at it, the first option does not properly handle pathfinding to a pose instead of a path, since it would just keep flipping the target pose each time. So, the "original target pose" solution is needed. Sorry for the confusion.

AJahueyM:
> Actually I think the second option I told you is the only one that would work. Now that I'm looking at it, the first option does not properly handle pathfinding to a pose instead of a path, since it would just keep flipping the target pose each time. So, the "original target pose" solution is needed. Sorry for the confusion.

Okay, it is the one I implemented I think, I just moved the initialization of everything to the Initialize

AJahueyM:
The original pose solution turned to be a lot more cleaner than what I previously did. I tested the changes on C++ and its working fine. I don't really have a way to test the Java or Python implementations.

PR #579, opened 2024-01-26
Add driver station warning if named command is not known

pjreiniger:
My kids had a capitalization problem with a named command and didn't know why their thing wasn't getting scheduled. By at least logging a warning, there is some insight to what is going wrong.

Alternatively, you can hard fail with an exception. I think it is bad that there can be a discrepancy that silently fails between what you see in the gui and what gets run on the robot.

### DIFF
--- a/pathplannerlib-python/pathplannerlib/auto.py
+++ b/pathplannerlib-python/pathplannerlib/auto.py
@@ -9,7 +9,7 @@ from .commands import FollowPathRamsete, FollowPathHolonomic, FollowPathLTV, Pat
     PathfindThenFollowPathLTV
 from .geometry_util import flipFieldPose
 import os
-from wpilib import getDeployDirectory, reportError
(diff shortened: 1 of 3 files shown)


mjansen4857:
I think reporting a warning or error makes sense over throwing since it allows for easily disabling certain markers for testing just by commenting out the lines that register them, without having to delete them from the path.

Can you add the same warning for python as well. Relevant line is [here](https://github.com/mjansen4857/pathplanner/blob/0384490c7f7bb623b1db549abb0e2761510f5cca/pathplannerlib-python/pathplannerlib/auto.py#L54). You'd need to add `reportWarning` to the import [here](https://github.com/mjansen4857/pathplanner/blob/0384490c7f7bb623b1db549abb0e2761510f5cca/pathplannerlib-python/pathplannerlib/auto.py#L12)

PR #603, opened 2024-02-08
Fix paths with events on C++

AJahueyM:
The latest updates that changed event markers to be time based caused paths with events to crash on C++.  
When doing:
```
m_untriggeredEvents.insert(m_untriggeredEvents.end(),
  m_generatedTrajectory.getEventCommands().begin(),
  m_generatedTrajectory.getEventCommands().end());
```

A new vector is created each time m_generatedTrajectory.getEventCommands() is called, meaning that the iterators are not linked. I solved this by storing the eventCommands in a different variable and then getting the begin and end iterators from it. Another solution could maybe be changing PathPlannerTrajectory::getEventCommands to return a constant reference to the events.


mjansen4857:
Thanks. Will replace the 2023.3.3 update I just did today after this is merged

PR #619, opened 2024-02-27
Python Event Marker Requirements Matches Java

### LINKED ISSUE #617: Python library checks requirements differently from Java
https://github.com/mjansen4857/pathplanner/blob/0cc0d9c9e03bb759acc52608def568ee94cc78ff/pathplannerlib-python/pathplannerlib/commands.py#L67

https://github.com/mjansen4857/pathplanner/blob/0cc0d9c9e03bb759acc52608def568ee94cc78ff/pathplannerlib/src/main/java/com/pathplanner/lib/commands/FollowPathCommand.java#L77

https://www.chiefdelphi.com/t/pathplanner-error-when-adding-requirements-to-command/456352?u=bovlb

Comparing these two implementations, the Python version checks against `self.getRequirements()`, which is all requirements gathered so far, whereas the Java version only checks against drive requirements.  Given the text of the error message, I believe the Java version is right here.  Perhaps this line should read:
```
for req in requirements:
```

PickleFace5:
Fixes #617 
When multiple event markers used the same subsystem, a RuntimeError is thrown, since the intake subsystem would be added into the current requirements, which tricked the code into believing it was a drive subsystem.

### DIFF
--- a/pathplannerlib-python/pathplannerlib/commands.py
+++ b/pathplannerlib-python/pathplannerlib/commands.py
@@ -59,13 +59,14 @@ class FollowPathCommand(Command):
         self._replanningConfig = replanning_config
         self._shouldFlipPath = should_flip_path
 
-        self.addRequirements(*requirements)
+        self.driveRequirements = requirements
+        self.addRequirements(*self.driveRequirements)
 
         for marker in self._originalPath.getEventMarkers():
             reqs = marker.command.getRequirements()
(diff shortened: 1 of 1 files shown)


mjansen4857:
#618 about 3 minutes too late lol

PickleFace5:
Darn, I'll beat you next time 😂

PR #639, opened 2024-03-13
Implemented buildAutoChooser() in Python.

jserra99:
Was already doing this for my team's robotpy code so I figured why not implement it here. As far as my testing extends, it works, I am not 100% sure if I implemented the docs correctly, however.

### DIFF
--- a/Writerside/topics/pplib-Build-an-Auto.md
+++ b/Writerside/topics/pplib-Build-an-Auto.md
@@ -432,7 +432,7 @@ class RobotContainer:
 
 > **Note**
 >
-> This feature is only available in the Java version of PathPlannerLib
+> This feature is only available in the Java and Python versions of PathPlannerLib
 >
 {style="note"}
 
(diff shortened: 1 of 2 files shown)


mjansen4857:
The documentation part should be changed to have tabs for different languages to be consistent with the rest of the examples. You should be able to copy one of those examples, remove the C++ tab, and change the code part. Let me know if you need any help with that.

jserra99:
Done. Didn't catch that the first time around, was wondering how the switching worked.

Writerside/topics/pplib-Build-an-Auto.md
```diff
@@ -471,4 +474,29 @@ public class RobotContainer {
...
+        SmartDashboard.putData("Auto Chooser", self.autoChooser)
+    
+    def getAutonomousCommand():
+        return self.autoChooser.getSelected()
+```
+
+</tab>
+</tabs>
+
+</snippet>
```
mjansen4857, line 502: I believe this is a stray closing tag
```suggestion

```

mjansen4857 (changes requested):

Writerside/topics/pplib-Build-an-Auto.md
jserra99: Is it? That was there before, I simply moved it down. The md originally opened with <snippet id="build-an-auto"> and ended with </snippet> immediately after the java example code.

mjansen4857: Yeah you're right. Forgot that file is different

PR #640, opened 2024-03-13
Altered Naming Convention for Duplicates

Ammelll:
In my experience the current naming convention of "Copy of {AutoName}" is very inconvenient, primarily when sorting the Path/Auto names alphabetically as duplicating an auto will send it to a random place in the stack instead of right next to the Path/Auto being duplicated. I believe the changes I've proposed would address this.

### DIFF
--- a/lib/pages/project/project_page.dart
+++ b/lib/pages/project/project_page.dart
@@ -889,9 +889,17 @@ class _ProjectPageState extends State<ProjectPage> {
         for (PathPlannerPath path in _paths) {
           pathNames.add(path.name);
         }
-        String pathName = 'Copy of ${_paths[i].name}';
+        String pathName = _paths[i].name;
+        RegExp exp = RegExp(r'\(\d+\)');
+        String source = pathName.substring(pathName.length-3);
         while (pathNames.contains(pathName)) {
-          pathName = 'Copy of $pathName';
+          if(exp.hasMatch(source)){
+            RegExpMatch? match = exp.firstMatch(source);
+            int index = int.parse(match![0]!.substring(1,2))+1; 
+            pathName = '${pathName.substring(0,pathName.length-3)}($index)'; 
+          } else{
+          pathName = '$pathName (1)';
+          }
         }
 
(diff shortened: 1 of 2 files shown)


rjbell4:
I concur that something that appends a suffix would be preferred over something the prepends a prefix, as it would result in less scrolling when creating new paths or autos by duplicating something that exists and then modifying it.

mjansen4857:
Just checking in on the status of this as it still has failing tests and needs to be formatted. 

You can format by running `dart format .` in the root of the project directory.

mjansen4857:
It looks like the tests are failing because the actual naming logic is not working as expected. For example, when duplicating `Example Path (1)` it expects to see `Example Path (2)` but doesn't. I'm assuming its creating `Example Path (1) (1)`

lib/pages/project/project_page.dart
```diff
@@ -908,9 +908,17 @@ class _ProjectPageState extends State<ProjectPage> {
         for (PathPlannerPath path in _paths) {
           pathNames.add(path.name);
         }
-        String pathName = 'Copy of ${_paths[i].name}';
+        String pathName = _paths[i].name;
+        RegExp exp = RegExp(r'\(\d+\)');
+        String source = pathName.substring(pathName.length - 3);
         while (pathNames.contains(pathName)) {
-          pathName = 'Copy of $pathName';
+          if (exp.hasMatch(source)) {
```
mjansen4857, line 915: I'm assuming tests are failing because you're matching the regex against the substring with the `(x)` removed. I think this should be changed to match against the entire path name and then only do the substring if it has a match. Same thing for the autos.

mjansen4857: When fixing this, could you also add on to the duplication tests to duplicate them a third time. i.e. check for `Example Path (3)` just to be extra sure the increment works.

Ammelll:
@mjansen4857 anything holding this up?

Ammelll:
@mjansen4857 bump

PR #658, opened 2024-03-19
Corrected Field Position Flip

trindels:
Field Position Flipping should flip across the Y-axis instead of the X-axis.


mjansen4857:
The field position flipping maintains a blue alliance origin, so it is correct.

trindels:
> The field position flipping maintains a blue alliance origin, so it is correct.

So, if the goal is to maintain the blue alliance origin, documentation should include the recommendation that the pose_supplier should be configured to always maintain the blue alliance origin as well?

PR #670, opened 2024-03-31
Added buildAutoChooser to C++ lib

ThePhoenixFox:
Added a equivalent version of buildAutoChooser for cpp from java.
Using the std::filesystem, we can inspect the files on the drive.
Since SendableChooser doesn't allow for rvalue storage, we need to store the commands ourselves so the user will not get a nullptr when trying to access a command* from GetSelected().

### DIFF
--- a/Writerside/topics/pplib-Build-an-Auto.md
+++ b/Writerside/topics/pplib-Build-an-Auto.md
@@ -440,12 +440,6 @@ class RobotContainer:
 
 ## Create a SendableChooser with all autos in project
 
-> **Note**
->
-> This feature is only available in the Java and Python versions of PathPlannerLib
->
-{style="note"}
-
 After configuring the AutoBuilder, you have the option to build a SendableChooser that is automatically populated with
 every auto in the project.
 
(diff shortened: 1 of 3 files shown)


mjansen4857:
I'd recommend running builds locally to help find any issues a bit easier.

`./gradlew build`

You can also fix formatting issues with `./gradlew spotlessApply`

ThePhoenixFox:
Sorry, this is my first time doing this. When I tried to run gradlew on my device, I got errors saying that the settings couldn't be found.

mjansen4857:
I usually have pretty good luck with IntelliJ so you might want to give that a try. It has built in gradle support that usually takes care of setting up everything without any issues.

WPILib VSCode also works. The "Build robot code" option should do the same thing. Does that work for you? You have to also make sure you have the `pathplannerlib` folder open and not the root `pathplanner` repo folder or it won't work.

ThePhoenixFox:
There are two versions of getAutonomousCommand() in the Writerside wiki. The first one is the easy readable one while the second one creates a frc2::CommandPtr since that is what most example programs use for their default return type.
https://github.com/mjansen4857/pathplanner/blob/da4d22c8b1edd5ede5ae7ba3d6afb62f8c20eec7/Writerside/topics/pplib-Build-an-Auto.md?plain=1#L506-L514

PR #672, opened 2024-04-02
changed "getStaringPoseFromAutoFile" to non-misspelled version

ironon:
our programmers tore their hair out over this


mjansen4857:
Can you add back a version of the method that is misspelled that just calls the correct method, then deprecate it? If the typo is just fixed this can't be released without breaking the code of everyone that uses that method.

ironon:
added 👍 
good catch, I should have thought of that originally.

mjansen4857:
No longer relevant as of #714

PR #674, opened 2024-04-04
Add buildAutoChooserWithOptionsModifier 

KennanHunter:
I created a new method called `buildAutoChooserWithOptionsModifier`, that accepts a second parameter in the form of a lambda function that takes in a stream and returns a stream of PathPlannerAutos.

I also added documentation to filter these options based on if they included the text "comp" at the beginning of the name. Some other examples you could do would be override the name with `.setName`, filter depending on if you're red or blue, or use [`.sorted()`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#sorted--) on the stream to sort them arbitrarily before going on shuffle board.

I couldn't run spotless for some weird errors in my environment I need to figure out but the formatting should be good enough. I also manually wrote that documentation so no guarantees it actually works with Writerside 🤷.

P.S. pathplanners great, thanks for all the work you do :)

### DIFF
--- a/Writerside/topics/pplib-Build-an-Auto.md
+++ b/Writerside/topics/pplib-Build-an-Auto.md
@@ -510,4 +510,43 @@ class RobotContainer:
 </tab>
 </tabs>
 
+## Create a SendableChooser with certain autos in project
+
+> **Note**
+>
+> This feature is only available in the Java version of PathPlannerLib
+>
+{style="note"}
+
+You can use the buildAutoChooserWithOptionsModifier method to process the 
+autos before they are shown on shuffle board
+
+```java
+public class RobotContainer {
+  private final SendableChooser<Command> autoChooser;
+
+  public RobotContainer() {
+    // ...
+
+    // Check if currently attached to the FMS
+    boolean isCompetition = DriverStation.isFMSAttached();
+    
+    // Build an auto chooser. This will use Commands.none() as the default option.
+    // As an example, this will only show autos that start with "comp" while
+    // currently attached to FMS
+    autoChooser = AutoBuilder.buildAutoChooserWithOptionsModifier(
+      (stream) -> isCompetition
+        ? stream.filter(auto -> auto.getName().startsWith("comp")) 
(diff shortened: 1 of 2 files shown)


mjansen4857:
Well format bot is broken so that's unfortunate

KennanHunter:
It looks like the format bot is trying to branch origin/main and doesn't know how to merge it up onto the pr... not exactly sure how thats supposed to work.

KennanHunter:
It looks like the format bot is trying to branch origin/main and doesn't know how to merge it up onto the pr... not exactly sure how thats supposed to work.

mjansen4857:
Yeah. It seems to work for some people but not others for whatever reason.

mjansen4857:
That workflow dispatch change wouldn't work since the workflow is very specific to comments

KennanHunter:
still didn't work damn, lemme revert those didn't realize they would get put in this pr immediately

Writerside/topics/pplib-Build-an-Auto.md
```diff
@@ -510,4 +510,43 @@ class RobotContainer:
...
+
+```java
+public class RobotContainer {
+  private final SendableChooser<Command> autoChooser;
+
+  public RobotContainer() {
+    // ...
+
+    // Check if currently attached to the FMS
+    boolean isCompetition = DriverStation.isFMSAttached();
```
mjansen4857, line 532: Bit of a nitpick but the example here likely wouldn't work since it is possible for code to boot up before connecting to the FMS. Perhaps just setting the boolean to true would be fine?

mjansen4857 (changes requested):

Writerside/topics/pplib-Build-an-Auto.md
KennanHunter: Thank you! Not something I would of thought of...

tbeadle:
@mjansen4857 any idea when a release will be cut with this change in it? Thanks!

PR #678, opened 2024-04-09
Fix typo

5438-Team-Account:
Noticed a small typo in the name of the method to get the starting pose from an auto file. This change will make it so the naming of the method reflects the naming in the docs.


5438-Team-Account:
nvm didn't realize someone already opened a pr

PR #679, opened 2024-04-12
Added filtering to buildAutoChooser for C++

ThePhoenixFox:
Added a new input for buildAutoChooser for C++ which allows filtering out autos based on their internal data like name or start position.
Default Option is guaranteed to be unfiltered to prevent nullptr. 
[GetSelected()](https://github.com/wpilibsuite/allwpilib/blob/74f648689ecd411f7529ca3f8ac9e7eec25758bc/wpilibc/src/main/native/include/frc/smartdashboard/SendableChooser.h#L72-L83) causes a default initialized Command* to nullptr.

Added buildAutoChooser with filtering to writerside wiki explaining how it works.

### DIFF
--- a/Writerside/topics/pplib-Build-an-Auto.md
+++ b/Writerside/topics/pplib-Build-an-Auto.md
@@ -539,4 +539,114 @@ class RobotContainer:
 </tab>
 </tabs>
 
+## Create a SendableChooser with certain autos in project
+
+> **Note**
+>
+> This feature is unavailable in the Python version of PathPlannerLib
+>
+{style="note"}
+
+You can use the buildAutoChooserWithOptionsModifier method to process the 
+autos before they are shown on shuffle board
+
+> **Warning**
+>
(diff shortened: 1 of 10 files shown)


ThePhoenixFox:
A note to add about the filtering function: the pointers are only valid if the function returns true. If a user want a do a shallow copy of the pointer, it needs to return true, otherwise they will get a nullptr because the function deletes unused commands. They must do a deep copy if they want to store it for later.

I don't know if you want to put this info in the API and wiki but it could be a edge case for some users.

mjansen4857:
Needs some conflicts to be resolved then I can merge this. It looks like its just conflicts with includes. You can accept the version from the current main branch.

ThePhoenixFox:
It been a while since I reviewed my code and I am just going to remove this pull request and make a new one since pathplanner has changed a bit.
I also want to address some of my inefficiencies which I brought into code such as always making new commands reference everytime a new frc::SendableChooser is requested. I want to get this done before my code needs to become backward capability.

PR #687, opened 2024-05-02
Modifications To PathPlanner

Midnight9385:
- hermite splines
- full log of telemetry


Midnight9385:
I need to add the tests folder back. I had removed it because it wouldn't compile and I didn't want it to affect the main app.

Midnight9385:
- updated the java lib to read the hermite splines
- I'm not in love with the replanning for hermite splines. It's a little more demanding then I would like it to be because of some edge cases. So for right now I just made it add points in the way it would be done if there were no bezier points to reference. Any suggestions or ideas would be appreciated.

Midnight9385:
Is there any specific reason the dependencies that are currently used are "outdated", is there any harm in updating them to the latest version?

PR #700, opened 2024-05-26
Replicate fix in commit: fa5020b for python lib (pathfinding starting…

roberttuttle:
… from/near goal position)

When migrating current season comp robot from Java to Python, we discovered in testing that the fix in this commit (https://github.com/mjansen4857/pathplanner/commit/fa5020b2fbc229bc87ea5ee1a0a9043a3744cd24) has not been ported over. (Note: we have not looked at any other newer commits to the main lib since the last python lib updates ... just this one that affects us directly in our autos).

### DIFF
--- a/pathplannerlib-python/pathplannerlib/commands.py
+++ b/pathplannerlib-python/pathplannerlib/commands.py
@@ -286,6 +286,8 @@ class PathfindingCommand(Command):
 
     _timeOffset: float = 0
 
+    _finish: bool = False
+
     _instances: int = 0
 
     def __init__(self, constraints: PathConstraints, pose_supplier: Callable[[], Pose2d],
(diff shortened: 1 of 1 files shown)


mjansen4857:
No, everything seems fine. I've just been mostly ignoring PRs/issues for a while to focus on finishing the trajectory rewrite. But this PR is simple enough to not really affect that so I'll get it merged.

roberttuttle:
Thank you! Was just about to start exploring how to build and install a local package based on our fork through RobotPy installer before our upcoming offseason event. This is super helpful and thanks for all the work you do on this library/project.

mjansen4857:
So sorry I had a lot going on and forgot about this. I'm about to merge #694 so I just replicated this in that so you won't have to deal with any merge conflicts

PR #725, opened 2024-09-22
Update Trajectory State to support Choreo moduleForces

Eaglegrant:
As seen in [Choreo's new TrajectoryState](https://sleipnirgroup.github.io/Choreo/api/choreolib/java/com/choreo/lib/ChoreoTrajectoryState.html), moduleForces are pre-calculated. 
This exposes those moduleForces, and updates the PathPlannerPath to make that possible.


mjansen4857:
PathPlannerTrajectory is being updated to calculate and store its own module forces via #724. Support for Choreo's updates will need to be added to fit with that implementation. So, I'll leave this PR alone for now until that gets ironed out.

mjansen4857:
After talking with the Choreo people a bit more it looks like Choreo's module forces are not going to be compatible with PPLib's motor torque feedforward stuff since those forces include the force due to friction. So, if anyone is going to use those fields, they will need to do so in a custom path following command. In that case, it makes more sense to just load the paths with ChoreoLib instead of PPLib so I don't think including these fields in PPLib makes a lot of sense.

mjansen4857:
Added in a different PR

PR #726, opened 2024-09-22
Added filtering for buildAutoSendable()

ThePhoenixFox:
SendableChooser references a map of all of the auto commands 
Filter provides access to auto command (const) and auto's path
Allows multiple SendableChooser
Allows commands to be updated to latest in pathplanner/auto directory (regenerateSendableReferences()) 
Provided wiki info for filtering

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -3,9 +3,6 @@ name: Build Documentation
 on:
   workflow_dispatch:
 
-  push:
-    branches: [main]
-
   pull_request:
 
 env:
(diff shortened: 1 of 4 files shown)


mjansen4857:
Out of curiosity, what's the use case for the `regenerateSendableReferences` method? Add new autos to the chooser if you manually copy them to the rio or something?

ThePhoenixFox:
Initially, it generates the map of all of the auto commands that are in "deploy/pathplanner/autos" which get used when someone generates a sendableChooser. The map gets referenced everytime a buildAutoChooser is called. 
If a team uploads new autos or modifies autos without deploying (like filezilla), they can use update the commands that are in the map to match the most recent in filesystem which the sendableChooser references.

Also got another commit, the directory filter example isn't correct.

mjansen4857:
Could you remove the 
```
push:
    branches: [main]
```
trigger for the `publish-docs.yml` actions workflow? I want to stop pushing docs updates unless the features in the docs are actually released. I would do it myself but don't want to make you merge main for such a minor change.

Issue #764, opened 2024-09-29
Have an option to "start late" a path


lidor51:
For example we would like to start late a path in case the actual starting position has changed. Might also improve PathfindingAndThenFollowPath command.

We would like to start late the path to align with the actual starting position of the robot.

Issue #698, opened 2024-05-12
Have an option to "end early" a path


lidor51:
For example we would like to end earlier a path in case we collected a game piece 1m/1sec earlier than planned.

We would like to end earlier the path to save time (and start rhe next path from this pose, using the appropriate replanning config).

lidor51:
In this case we might want to tirgger in order all remaining markers.

lidor51:
@mjansen4857 IMO it's super important feature for the future of autonomous, it'll allow better support in game-pieces vision based decisions. How it's aligned with this summer prioritization?

HenryLeC:
This is awkward. I just replied to you in #721 with my PR (#923), but it also allows for this, and it is one of the examples. You just need to add the FollowPath and WaitUntil to a Race group

lidor51:
My point here wasn't on how to technically end path early, as it's possible as you stated. The thing to consider when ending path early is that the next path will consider the new starting point location (that might be dynamic and changed based on where a game piece was intaked).

HenryLeC:
I did play around with this problem in simulation and the solution I’m going to try on a real robot is starting the next path in a location that you can pathfind to and then path finding to it before starting the path. The solution is not ideal because it requires runtime computing of the path so is therefore maybe nondeterministic, but pathfinding has been shown to work pretty well during the 2024 season. You will also be abruptly stoping the robot between the paths which is not ideal and may cause problems.

Currently there is no way to pathfindThenRunPath from the UI so once I get to test this I am going to add that to the UI and lib

lidor51:
I see some scenarios in which there are issues with your solution to patch pathfind to the next path (e.g., we need to run a shorter path than the original path, so pathfinding to the original start point will make our auto longer than needed), instead if we're too far from the next starting point, running pathfind to pose would be a general solution that can work. (Technically on the fly generation would also work but it's harder to implement a solution that works in the general case).

My suggestion:
 Implement Pathfind to **Pose** from UI, that need to have an adequate solution to trigger other subsystems commands during pathfinding (e.g., based on distance from start point or to end point).

PR #765, opened 2024-09-29
Support trimming an existing trajectory

lidor51:
This PR supports a trimming of a trajectory. Introduce an elegant way to start a trajectory in the middle (and also to terminate it earlier).

My original thought was to trim a PathPlannerPath, but it would be less user friendly as the params were WaypointPos instead of seconds.
**Let me know if you think it's better to add trim support to a path instead to a trajectory.**

#764
#698

### DIFF
--- a/pathplannerlib/src/main/java/com/pathplanner/lib/trajectory/PathPlannerTrajectory.java
+++ b/pathplannerlib/src/main/java/com/pathplanner/lib/trajectory/PathPlannerTrajectory.java
@@ -17,6 +17,7 @@ import edu.wpi.first.math.kinematics.SwerveModuleState;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /** Trajectory generated for a PathPlanner path */
 public class PathPlannerTrajectory {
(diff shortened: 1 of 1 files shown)


mjansen4857:
This isn't something that I think should be in the main library. Trimming trajectories like this can lead to some pretty bad path following performance very easily. If you cut out part of the trajectory in the beginning handling acceleration, the robot would be commanded to go much faster than it is capable of doing and the robot will fall behind for the whole trajectory.

This method is something that can be easily implemented in user code, and if you know what you're doing when using it then it can work fine, but if its not something I would recommend anyone to use because of the potential issues caused by it then it just probably shouldn't be in the library. I want to leave anything like this outside of just following the trajectory as-is up to the user to implement themselves so they are fully aware of what is happening.

lidor51:
If you can expose also getters for PathPlannerPath, it'll be appreciated, so advanced users can make modifications to the path without modifying the json object in runtime. Do you want me to open a feature request? @mjansen4857

mjansen4857:
What getters? I think everything needed should already be exposed.

lidor51:
Getter of number of waypoints (not path points) & getter of rotation targets (although can be extracted from iterating through path points), would be appreciated.

mjansen4857:
Yeah you should be able to trim the path purely off path points but all that stuff should be publicly accessible anyway so I’ll do that.

PR #784, opened 2024-10-02
add route reverse feature and add auto save when clone and reverse

ycy-0510:
### add route reverse feature: 
user and build route for alliances
### and add auto save when clone and reverse: 
use don’t need to edit before close app

### DIFF
--- a/lib/pages/home_page.dart
+++ b/lib/pages/home_page.dart
@@ -98,7 +98,7 @@ class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
 
       if (projectDir == null || !fs.directory(projectDir).existsSync()) {
         projectDir = await Navigator.push(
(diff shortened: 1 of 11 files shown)


mjansen4857:
So this looks like its for flipping paths to the other side of the field, which is not something that will be supported in the GUI. PathPlannerLib already takes care of flipping paths, which prevents the need to maintain a different path for each alliance.

ycy-0510:
Some teams maybe need to flip and adjust some value. They can use this feature.

mjansen4857:
In that case they can just duplicate the path and let PPLib flip it. I don't want to include any flipping logic in the GUI since it overcomplicates things like switching field images and I just wouldn't recommend people use it over the PPLib flipping anyway.

PR #862, opened 2024-10-13
Style changes to buildAutoChooser

ThePhoenixFox:
Changed filtering function from `const PathPlannerAuto* const`-> `const PathPlannerAuto&`: Easier syntax, reference act like varible instead of pointer
`const_cast<>` can still change the const-ness of `PathPlannerAuto&` which allows them to edit directly (I despise this feature, however there is no copy constructor for `PathPlannerAuto` due to `unique_ptr`)
Replace `dynamic_cast<>` to `static_cast<>` since we know relationship at compile time 
Set `filter` as first parameter and `defaultAutoName` as second (which is defaulted), which prevents confusion as to why there is an empty string input (in user code)
Separated `buildAutoChooser` into `buildAutoChooser`, `buildAutoChooserFilter`, and `buildAutoChooserFilterPath` for different degrees of concern and to mirror java version more neatly (allows people to change from both languages)
Updated function descriptions

Degrees of concern for `buildAutoChooser`
* Person who just wants to get a `SendableChooser` of all PathPlannerAutos just needs to worry about `buildAutoChooser`
* Person who wants to `filter` out some PathPlannerAutos just needs to worry about `buildAutoChooserFilter`
* Person who wants to `filter` out some PathPlannerAutos based on filepath (`std::filesystem::path` libary) needs to worry about `buildAutoChooserFilterPath` 

I know I have made a lot of changes to this feature however I just want it to be right before backward compatibility requirements happen. I am probably not going to edit this any further as it performs all the roles it need to Java version while maintaining C++ style.
Good luck rolling out for the 2025 season btw👍

### DIFF
--- a/Writerside/topics/pplib-Build-an-Auto.md
+++ b/Writerside/topics/pplib-Build-an-Auto.md
@@ -540,33 +540,29 @@ RobotContainer::RobotContainer() {
   // Build an auto chooser. This will use frc2::cmd::None() as the default option.
   // As an example, this will only show autos that start with "comp" while at
   // competition as defined by the programmer
-  autoChooser = AutoBuilder::buildAutoChooser(
-    "", // If empty it will choose frc2::cmd::None()
-    [&isCompetition](const PathPlannerAuto *const autoCommand,
-            std::filesystem::path autoPath)
+  autoChooser = AutoBuilder::buildAutoChooserFilter(
+    [&isCompetition](const PathPlannerAuto& autoCommand)
     {
-      return isCompetition ? autoCommand->GetName().starts_with("comp") : true;
+      return isCompetition ? autoCommand.GetName().starts_with("comp") : true;
     }
   );
 
   // Another option that allows you to specify the default auto by its name
   /*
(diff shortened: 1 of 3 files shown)


mjansen4857:
Tiniest nitpick ever but it looks like a few of the function descriptions have a typo in “recurively”

ThePhoenixFox:
Yeah you change it to "recursively". That is what happens when spell check isn't there for you.

ThePhoenixFox:
PPLib-macOS is commented out in in pplib-ci.yml so that is why it not finishing. I do not know if that is intentional but that is why the workflow isn't completing (since it also marked as require).
https://github.com/mjansen4857/pathplanner/blob/4e74a202d6332f3025a607dc0f58e6a8ab5dc304/.github/workflows/pplib-ci.yml#L172C1-L174C29

mjansen4857:
Yeah I meant to merge this earlier and forgot. macOS is disabled because it fails to build with 2024 WPILib for whatever reason. Works fine with 2025 so I just disabled it temporarily.

PR #883, opened 2024-10-23
Fix `FlippingUtil` behavior with `kMirrored` robot-relative forces and `kRotational` field rotation

bhall-ctre:
1. The robot-relative feedforward Y components need to be inverted in `kMirrored` mode, as left and right are flipped.
2. In `kRotational` mode, the field rotation needs to be rotated by 180 degrees, not subtracted from 180 degrees. For example, a 90 degree rotation should become -90 after rotating, but stay 90 degrees after mirroring.

I wasn't sure what the best way to implement 1 is API-wise, so for now I implemented two new functions `flipFeedforwardXs` and `flipFeedforwardYs` to handle the two components. Feel free to rename the functions or try to merge them into the existing `flipFeedforwards` function

### DIFF
--- a/pathplannerlib-python/pathplannerlib/util.py
+++ b/pathplannerlib-python/pathplannerlib/util.py
@@ -77,8 +77,8 @@ class DriveFeedforwards:
             FlippingUtil.flipFeedforwards(self.accelerationsMPS),
             FlippingUtil.flipFeedforwards(self.forcesNewtons),
             FlippingUtil.flipFeedforwards(self.torqueCurrentsAmps),
-            FlippingUtil.flipFeedforwards(self.robotRelativeForcesXNewtons),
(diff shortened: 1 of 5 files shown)


pathplannerlib-python/pathplannerlib/util.py
```diff
@@ -154,6 +157,31 @@ def flipFeedforwards(feedforwards: List[float]) -> List[float]:
                 return [feedforwards[1], feedforwards[0]]
         return feedforwards
 
+    @classmethod
```
mjansen4857, line 160: Out of curiosity, why `@classmethod` here instead of `@staticmethod`?

bhall-ctre: I did that to use `clazz.flipFeedforwards()` in the body, but you can replace it with `@staticmethod` and `FlippingUtil.flipFeedforwards()` instead if that's what you standardize on.

PR #911, opened 2024-11-18
Add Empty Translation check for forceVec

PickleFace5:
On robot initialization, the console output would have the following warning occasionally:

```Error at `anonymous-namespace'::WPILibMathShared::ReportErrorV: Error: x and y components of Rotation2d are zero```

After some print debugging, I found this was caused by forceVec being an empty Translation2d. This simply adds a check for the forceVec Translation2d to avoid sending warnings to the console.

I'm unaware if this issue is present in other languages.

### DIFF
--- a/pathplannerlib-python/pathplannerlib/trajectory.py
+++ b/pathplannerlib-python/pathplannerlib/trajectory.py
@@ -497,7 +497,10 @@ def _forwardAccelPass(states: List[PathPlannerTrajectoryState], config: RobotCon
 
             # Calculate the torque this module will apply to the robot
             angleToModule = (state.moduleStates[m].fieldPos - state.pose.translation()).angle()
-            theta = forceVec.angle() - angleToModule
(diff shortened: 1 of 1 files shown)


PickleFace5:
I have added checks to Java and C++ with the recommended implementation.

mjansen4857:
Looks like build is failing just for some unit stuff. You just need to add an extra `()` after the C++ `Norm()` calls. So `translation.Norm()() <= 1E-6`

To fix formatting, you can do `./gradlew spotlessApply`

mjansen4857:
Also, can you replace usages of `new Rotation2d()` with `Rotation2d.kZero` in Java? Helps reduce memory pressure a bit.

PickleFace5:
@mjansen4857 added your requested changes. Hopefully that was the only syntax error for C++, I'm having issues with VS Code finding the header files. 😅

PR #917, opened 2024-11-25
Add Swerve Setpoint Generator to PPLib Python

PickleFace5:
I have translated the SwerveSetpointGenerator and such to Python, however there’s a few more things I plan to add:

- [x] ~~Additional Kinematics support (currently only works with 4 modules)~~ A fair bit of PathPlanner currently only works with 4 modules, so this won't be added.
- [x] Additional init constructor

I also need to test to ensure I translated it correctly.


mjansen4857:
I don't think additional kinematics is required at the moment, as various places in the library for all languages currently only support using 4 modules. This is mainly due to path flipping, which can only handle properly flipping feedforwards if the modules are in FL, FR, BL, BR order, which is only possible when using 4.

PickleFace5:
I figured that was the case, but I kept is as a possible addition. I'll probably rewrite a bit for it to solely work with 4 modules.

PickleFace5:
After fixing all the type errors and adding the additional constructors, I seem to be having a logic error with the drive motor calculations. The steer motors seem to be working correctly, but the drive motors output nearly nothing. This could be from my implementation, but I'll have to double check my translation when I get the chance sometime soon.

PickleFace5:
Still no luck 😔

PickleFace5:
I forgot 2 parenthesis, but after adding them back the setpoint generator works! All that's left to do is to update the documentation, but I'll mark this PR as ready for review.

pathplannerlib-python/pathplannerlib/util/__init__.py
```diff
@@ -0,0 +1,736 @@
...
+        return float('inf')
+    return sign * (ab * bc * ac) / (4 * area)
+
+
+def decimal_range(start: float, stop: float, increment: float):
+    while start < stop and not math.isclose(start, stop):
+        yield start
+        start += increment
+
+"""
```
mjansen4857, line 294: Think this should probably be removed

Writerside/topics/pplib-Swerve-Setpoint-Generator.md
```diff
@@ -70,3 +73,51 @@ public class SwerveSubsystem extends Subsystem {
...
+
+    self.setpointGenerator = SwerveSetpointGenerator(
+      config, # The robot configuration. This is the same config used for generating trajectories and running path following commands.
+      rotationsToRadians(10.0) # The max rotation velocity of a swerve module in radians per second. This should probably be stored in your Constants file
+    )
+
+    # Initialize the previous setpoint to the robot's current speeds & module states
+    currentSpeeds = self.getCurrentSpeeds() # Method to get current robot-relative chassis speeds
+    currentStates = self.getCurrentModuleStates() # Method to get the current swerve module states
+    self.previousSetpoint = SwerveSetpoint(currentSpeeds, currentStates, config.numModules)
```
mjansen4857, line 103: This needs to pass in feedforward, not the number of modules.

```suggestion
    self.previousSetpoint = SwerveSetpoint(currentSpeeds, currentStates, DriveFeedforwards.zeros(config.numModules))
```

PickleFace5:
Should be good now.

mjansen4857:
Could you duplicate the check from #927 in this?

Issue #609, opened 2024-02-17
Add an Either Command to the pathplanner events


mmilunicmobile:
Something that my team is working on is conditional pathing based on whether we have a game piece in the robot. We are currently making named commands that follow one sequence of paths if we have a piece and a different set if we do not see a piece. This saves time not having to go back.

A solution to this would be including an "either" group in addition to the current "sequential", "parallel", etc. groups. This could allow for much more dynamic branching in auto. Boolean suppliers could be registered with the named commands static class as well.

NickAnastasoff:
id love to see this! i came across conditional commands which could help https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/ConditionalCommand.html

mmilunicmobile:
After thinking about this for a bit, a hack that you could kinda run would be using a sequential command group inside of a race command group with one other command. The other command could be something like "hasPiece", and when the command starts, it could immediately check if there is a piece, and either end immediately if there is no piece, or continue forever if there is one. This has the effect of the items in the sequential command group only running if there is a piece. You could then make a "!hasPiece" that does the opposite and put that in a race sequence immediately after. This would give you an equivalent to the Either. Since very very few teams would ever use this, and adding the ability to register boolean suppliers is a huge undertaking, I think that this solution would work for now, and we're probably gonna try and see if this works.

mmilunicmobile:
This is an example of the sort of composition @NickAnastasoff:

<img width="2560" alt="Screenshot 2024-03-21 at 10 39 33 PM" src="https://github.com/mjansen4857/pathplanner/assets/62996888/784c88db-7962-4bfe-974d-2643a83f7429">

NickAnastasoff:
I think at this point it would make more sense to just create the autos in code with the PathPlanner.fromPathFile function, although that is a cool way to do it!

Issue #721, opened 2024-09-14
Support branching autos from the UI


lidor51:
Defining autonomous as a prioritized decision tree, each node can be either a path or a decision tree of its own.
Doing this from the UI would allow user-friendly way to define complex autos.

![image](https://github.com/user-attachments/assets/6ccc7b01-a25b-4d9f-806f-5b9441a1e426)

Each branch would have a condition, conditions would be checked from left to right (this is why it's a "prioritized" decision tree). We would like to make the paths (\w their conditions?) draggable.
Conditions should be reusable for a user-friendly experience.
If neither of the conditions are met, the auto finished.

lidor51:
I suggest a vertical scrolling in this window, and scrolling/right-left buttons horizontally.
Alongside keyboards arrows navigation in the tree.

mjansen4857:
I think this would be useful, but this is a very large and time-consuming feature to implement. So, I'm not currently targeting it for 2025 unless all remaining tasks for 2025 are done with enough time left to properly implement it before the end of beta.

lidor51:
Indeed a big feature, I suggest to keep it open to a future iteration.

HenryLeC:
This isn't exactly what you were thinking but this is achievable with Conditional Command groups #923 
@lidor51 Is there anything you can think of that a directed graph could achieve that my pull request can't besides being cleaner?

lidor51:
Hey that's cool! Besides the different visualisation, it seems like there're a few differences from my suggestion.

I have a few questions on the screenshot you put:
![17335932707358633460928559814498](https://github.com/user-attachments/assets/58812644-5490-4d16-aea2-c86792ca6a6f)

1. Is 'Shoot from 1' is "on true" & 'Sweep line' is "on false"?
If so, not sure if this is a self explainatory experience.

2. Is it possible to add conditional group inside another conditional group's onTrue/onFalse?

3. IMO it's a very important feature being able to save a named sub-branch so you don't need to duplicate and re-define it again in every auto. Alternatively, a more general approach is supporting "sub-autos" (practically, being able to put an auto as a component of another auto).

HenryLeC:
1. Yes, thats exactly what it is. It's the same format as the code, and it also explains it in a tooltip if you hover over the group name. The docs will also be helpful for this
2. Yes, that is possible. You would just need to add a conditional command group into the group for onTrue or onFalse. (A conditional command group is not a command group in the pathplanner UI)
[...]

PR #923, opened 2024-12-06
Conditional Command group and Wait Until with Named Conditionals

HenryLeC:
Conditional command groups are interesting additions to Autos that will allow for conditionally changing pathing if a game piece was not picked up for example.
Wait until commands will allow for running a path until some condition is met. This could be used for running a specific path along a line of gamepieces while trying to pick them up until a gamepiece is detected in the robot.

NamedConditions are very similar to NamedCommands in the GUI and libraries

Example auto mode using wait until and conditional group
image

### DIFF
--- a/lib/auto/pathplanner_auto.dart
+++ b/lib/auto/pathplanner_auto.dart
@@ -2,6 +2,7 @@ import 'dart:convert';
 
 import 'package:file/file.dart';
 import 'package:path/path.dart';
+import 'package:pathplanner/commands/conditional_command_group.dart';
 import 'package:pathplanner/commands/named_command.dart';
 import 'package:pathplanner/commands/command.dart';
 import 'package:pathplanner/commands/command_groups.dart';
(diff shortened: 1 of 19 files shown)


HenryLeC:
Fixes #609

mjansen4857:
Sorry for just leaving this open for a while, but this is something I've decided a while ago that I would not add to PathPlanner/PPLib.

The command group building stuff is really designed for just doing simple/basic stuff, making it easier for teams with little to no auto experience to put together a simple auto. Conditional stuff is more complex functionality that would pretty much only be used by teams capable of creating auto command groups in code anyways. Plus, having that capability in the GUI doesn't really make it any easier to use conditionals, in some cases making it more difficult, as you have to add a named conditional, build the conditional command in the gui, then you have to keep track of the separate implementations of the conditional/group in the code/gui, and it becomes super difficult to debug.

Basically, I think its just a whole lot easier/safer to build conditional commands/autos in code. I wouldn't ever recommend someone to use something like this over just manually building it in code, so adding it to the library and having to maintain it going forward just doesn't make sense.

For #721 I have ideas/plans on how to handle it sort of like a behavior tree, using a blackboard that the user can control in code instead of using a "NamedConditionals" setup. But, this can't be extended to happen within command groups without some significant library restructuring and hacking to get it to work. So, conditional stuff will probably just live in a level above the auto command groups. However, I think this will mostly accomplish what this PR is trying to achieve anyways, since the tree would be able to satisfy the use case for the "conditional group", and conditional waits and stuff like that can always just be bound to a named command.

Anyways, anything significant in respect to any sort of conditional features would not be added for 2025 anyways, as its just too significant a change to make this close to kickoff.

PR #926, opened 2024-12-11
Add Swerve Setpoint Generator to PPLib C++

Luis-Leyva:
I translated the swervepoint generator to C++, but documentation still needs to be updated but just wanted to create the PR in case there is something wrong with the translation

- [x] Update Documentation.
- [x] Test on Simulation

### DIFF
--- a/pathplannerlib/src/main/native/include/pathplanner/lib/util/swerve/SwerveSetpointGenerator.h
+++ b/pathplannerlib/src/main/native/include/pathplanner/lib/util/swerve/SwerveSetpointGenerator.h
@@ -97,42 +97,23 @@ public:
 
 private:
 	double kEpsilon = 1e-8;
-	int MAX_STEER_ITERATIONS = 8;
-	int MAX_DRIVE_ITERATIONS = 10;
 
 	RobotConfig m_robotConfig;
 	units::radians_per_second_t maxSteerVelocity;
 	units::volt_t brownoutVoltage;
-	using Function2d = std::function<double(double, double)>;
 
-	/**
-	 * Find the root of the generic 2D parametric function 'func' using the regula falsi technique.
-	 * This is a pretty naive way to do root finding, but it's usually faster than simple bisection
-	 * while being robust in ways that e.g. the Newton-Raphson method isn't.
-	 *
-	 * @param func The Function2d to take the root of.
-	 * @param x_0 x value of the lower bracket.
-	 * @param y_0 y value of the lower bracket.
-	 * @param f_0 value of 'func' at x_0, y_0 (passed in by caller to save a call to 'func' during
-	 *     recursion)
-	 * @param x_1 x value of the upper bracket.
-	 * @param y_1 y value of the upper bracket.
-	 * @param f_1 value of 'func' at x_1, y_1 (passed in by caller to save a call to 'func' during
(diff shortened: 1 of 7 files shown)


mjansen4857:
Could you duplicate the check from #927 in this?

Luis-Leyva:
Done. Probably won't pass any tests due to wpilibsuite/allwpilib#7549, but probably you would like to fix that on another PR to keep things clear. Also I'll update the docs until the python implementation PR is merged to prevent a merge conflict.

mjansen4857:
Tests should be able to run now if you merge in main. Feel free to update docs whenever, the merge conflict would probably be very minor since it would just be separate tabs + removing the note about being Java exclusive. The python PR still needs that 927 check so whichever one is done first will be merged first.

Has this translation actually been run/tested in sim at least yet?

pathplannerlib/src/main/native/include/pathplanner/lib/util/swerve/SwerveSetpointGenerator.h
```diff
@@ -0,0 +1,128 @@
...
+	SwerveSetpointGenerator();
+
+	/**
+	 * Create a new swerve setpoint generator
+	 *
+	 * @param config The robot configuration
+	 * @param maxSteerVelocity The maximum rotation velocity of a swerve module, in radians
+	 *     per second
+	 */
+	SwerveSetpointGenerator(RobotConfig *config,
```
mjansen4857, line 33: Why take in RobotConfig as a pointer? I believe the rest of the library uses `const &` when dealing with config.

Luis-Leyva:
I have not tested it yet, will be working on that in the next couple of days, I'll let you know when it's done.

pathplannerlib/src/main/native/include/pathplanner/lib/util/swerve/SwerveSetpointGenerator.h
Luis-Leyva: Right, that's an unnecessary pointer right there. Just a question due to the RobotConfig not having a default constructor, should I pass it to the generateSetpoint function rather than having it as an attribute and pass it through the constructor of the class.

mjansen4857: You could probably just add a default constructor to RobotConfig

mjansen4857:
When you get a chance, can you add the changes from #930 as well?

Luis-Leyva:
I updated the docs and added the changed from #930, still need to test it, hope to get it done during the weekend.

mjansen4857:
Got another change to bring over: #948

mjansen4857:
Hopefully the last one: #944

Luis-Leyva:
A quick update on this, I've been testing it with my own code and CTRE Swerve Drive and found out that themaxSteerVelocity on the generator has to be set up around 90 to 100 rad/s to allow the robot to move but that just causes some weird behaviors due to the high value of the steering speed. I've tried to debug it but no luck yet, will update when I find the issue. Also not sure if that is caused by something else.

Luis-Leyva:
Should be good now, I was totally confusing the units, just made the change to match Java and Python.

PR #931, opened 2024-12-16
[Swerve Setpoint Generator] Limit max speed when voltage sags

TheComputer314:
This prevents an edge case where the torque limit breaks when the desired velocity is not possible for the motor given the current max supply voltage.

image


mjansen4857:
These changes would allow the setpoint generator to generate setpoints that are not attainable for the robot. Desaturating the modules based on the voltage would limit the speed to a maximum that it can't necessarily decelerate to because the desaturation does not take the motor torque/wheel friction into account. The ability for the motor torque to be negative is also important because that is what will cause the robot to slow down when the drive motor is moving faster than the battery voltage would allow.

Additionally, #930 sort of supersedes this PR by allowing the user to specify a static voltage to avoid the behavior of using a dynamic voltage.

mjansen4857:
I'm also not really sure what the graphs you posted are showing, so if I am missing some other issue that still persists, let me know.

TheComputer314:
Sorry for not adding an explanation to the graphs.

The first graph shows the setpoint velocity making a discontinuous jump when voltage sags below 12 volts, as the drive acceleration constraint stops taking effect due to `max_vel_step` reaching exactly 0, letting the other constraints take over, which ignores the voltage limited max speed.

The second graph can be ignored, as it was implementation error in a separate modification I made.

> The ability for the motor torque to be negative is also important because that is what will cause the robot to slow down when the drive motor is moving faster than the battery voltage would allow.

From my testing in simulation with these changes, the setpoint can still decelerate when the voltage sags, even in a unreasonable scenario of dropping 4 volts in 1 tick.

image

In the above image, I have a trigger bound such that when the button is pressed, the supply voltage fed into the generator is subtracted by 4. It's initially held, then released, then held again. Blue is setpoint, yellow is actual. If accounting for battery internal capacitance and a less extreme voltage sag, the deceleration is more gradual.

mjansen4857:
Ok so this will need to be updated a bit for some changes made to the generator, and needs the same changes made to the new python version. I can help with translation if needed.

TheComputer314:
What autoformatter does PPLib Python use?

mjansen4857:
I haven’t set one up yet. I just typically use the Code -> Format Code option in PyCharm. If you don’t use that then don’t worry about it.

mjansen4857:
So it looks like the python version is missing the current draw limits. This will also need to be ported to the C++ version as well which was just merged

PR #964, opened 2025-01-04
Add Alerts to RobotConfig

JosephTLockwood:
This pull request adds the ability to check whether hard-coded RobotConfig matches the configuration in GUI.

### DIFF
--- a/pathplannerlib/src/main/java/com/pathplanner/lib/config/RobotConfig.java
+++ b/pathplannerlib/src/main/java/com/pathplanner/lib/config/RobotConfig.java
@@ -325,48 +325,44 @@ public class RobotConfig {
       new Alert("Module Location Mismatch", AlertType.kError);
 
   /**
-   * Checks if this configuration differs from the GUI configuration. Loads the GUI config and
-   * compares all properties, setting alerts for any mismatches.
+   * Checks if this configuration matches the GUI configuration. Loads the GUI config and compares
+   * all properties, setting alerts for any mismatches.
    *
-   * @return true if any configuration is invalid or GUI config cannot be loaded, false if all match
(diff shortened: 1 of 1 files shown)


mjansen4857:
Could you invert the methods (and their names) to return true if the config is valid instead of false? Its pretty common for a false return value to indicate an error so I'd like to stick to that.

JosephTLockwood:
I don't have a problem with that. That is how I originally had them. I switched them because they would return the opposite of what you set your Alert to, which seemed counterintuitive. If there are going to be test cases, then would you want them to return the same? 

It's totally up to you. I'm good with either.

mjansen4857:
I think its better to have true be a "normal" return value and then the user code/tests can react to false as an error. The alerts and the method return value or sort of separate things so having different values there isn't a huge deal.

pathplannerlib/src/main/java/com/pathplanner/lib/config/RobotConfig.java
```diff
@@ -304,4 +306,171 @@ public static RobotConfig fromGUISettings() throws IOException, ParseException {
...
+      guiConfig = RobotConfig.fromGUISettings();
+    } catch (IOException | ParseException e) {
+      BAD_GUI_CONFIG.set(true);
+      return false;
+    }
+
+    return validatePhysicalProperties(guiConfig)
+        && validateDriveSystem(guiConfig)
+        && validateWheelProperties(guiConfig)
+        && validateModuleLocations(guiConfig);
```
mjansen4857, line 346: If one of these methods fails, all methods afterwards would not be run, meaning that alerts that should be set might not actually get set.

PR #990, opened 2025-01-09
Add annotated field image

ArchdukeTim:
I think many teams will find it helpful to use this annotated image when making paths.

It follows the FMS labels provided in the manual, so it should be accepted as a common labeling scheme


mjansen4857:
Could you change the default field image on line 26 to FieldImage.official(OfficialField.reefscape) so that the default doesn’t change? I should have done this earlier anyway but never did.

PR #995, opened 2025-01-10
Fix Reefscape annotated field not loading

Gold872:
Also changed the name in the drop-down menu to make it easier to differentiate


mjansen4857:
Just going to replace the builds on the release with a new build instead of a new release. Will take a few minutes

Gold872:
> Just going to replace the builds on the release with a new build instead of a new release. Will take a few minutes

I've done this more times than I'd like to admit, most recently was at 3:30 am today

Issue #1007, opened 2025-01-15
IdealStartingState Typo


codydg:
**Describe the bug**
There is a typo in the Javadoc for IdealStartingState. It mentions "when finishing a path" which appears to be a copy-paste error from GoalEndState.

**To Reproduce**
N/A

**Expected behavior**
N/A

**Screenshots**
[Line 11](https://github.com/mjansen4857/pathplanner/blob/main/pathplannerlib/src/main/java/com/pathplanner/lib/path/IdealStartingState.java#L11) and [Line 18](https://github.com/mjansen4857/pathplanner/blob/main/pathplannerlib/src/main/java/com/pathplanner/lib/path/IdealStartingState.java#L18) have the typo.

**Versions: (please complete the following information):**
 - OS: N/A
 - GUI Version: N/A
 - PPLib Version: 2025.2.1
 - PPLib Language: Java

**Additional context**
The C++ documentation does not have this typo.

PR #1008, opened 2025-01-15
Fix typo in IdealStartingState Javadoc (#1007)

codydg:
I believe I've fixed the typo.

If there could be more clarification added, I'd be interested. I find the term "Ideal" a bit vague. Does it mean the "expected" starting state? That would make sense to me since I'm using it for on-the-fly generation, but I'm not sure if there are other use cases of the class.

### DIFF
--- a/pathplannerlib-python/pathplannerlib/path.py
+++ b/pathplannerlib-python/pathplannerlib/path.py
@@ -109,7 +109,7 @@ class GoalEndState:
 @dataclass(frozen=True)
 class IdealStartingState:
     """
-    Describes the ideal starting state of the robot when finishing a path
+    Describes the ideal starting state of the robot when starting a path
 
     Args:
         velocity (float): The ideal starting velocity (M/S)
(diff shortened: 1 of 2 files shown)


mjansen4857:
It is basically an expected starting state, but not required. It's basically only used for GUI trajectories to pre-generate the "ideal" trajectory that will be followed if the robot's starting state matches. Otherwise, it will generate a new trajectory to match its actual state. The IdealStartingState is not needed for on-the-fly trajectories, since those will need to be generated anyways.

If you want to add a note about that, feel free.

mjansen4857:
Also, any changes to the javadoc probably need to be made to c++ and python docs as well.

codydg:
> The IdealStartingState is not needed for on-the-fly trajectories, since those will need to be generated anyways.

When I construct the `PathPlannerPath`, doesn't it make the time-based trajectory at that moment? I may have a core misunderstanding of how the generation works, but I'll give an example:

Generate a Path from <0, 0> to <1, 0> to <1, 10> (forwards 1, then left 10 units)
If I tell it my initial velocity is 10/s, I'd expect a different trajectory than if I said it's zero.

If I tell it null, it has already generated the path so I'm not sure how it would account for the 'sudden finding' at the start that I'm actually moving at 10/s.

> Also, any changes to the javadoc probably need to be made to c++ and python docs as well.

I did check the C++ documentation which just says "Create a new ideal starting state", meaning no typo was there.
I had forgotten about Python but I just checked and it had the same typo as Java so I've now pushed a fix.

mjansen4857:
The trajectory does not get generated until a path following command starts.

PR #1045, opened 2025-01-29
Add path mirroring for C++

r4stered:
This mirrors the functionality added in #987 for the C++ version of pathplannerlib


r4stered:
Not sure what is going on with CI. Seems like a network error not related to the changes

Issue #1056, opened 2025-02-09
On-The-Fly Path Generation When Already at Finished Pose Crashes Robot Code


blaze-developer:
When trying to follow an on-the-fly path created like this,

```java
public static Command straightLinePathCmd(Pose2d targetPose) {
    PathConstraints constraints = new PathConstraints(
      4.0,
      4.0,
      DegreesPerSecond.of(540).in(RadiansPerSecond),
      DegreesPerSecondPerSecond.of(720).in(RadiansPerSecondPerSecond)
    );

    List<Waypoint> waypoints = PathPlannerPath.waypointsFromPoses(
      AutoBuilder.getCurrentPose(),
      targetPose
    );

    PathPlannerPath path = new PathPlannerPath(
      waypoints,
      constraints,
      null,
      new GoalEndState(0.0, targetPose.getRotation())
    );

    path.preventFlipping = true;

    return AutoBuilder.followPath(path);
}
```

It normally works fine. However, when the robot is already at its final position, like when a driver accidentally presses the line-up button again after already lining up, the code crashes.

I tried checking if the list of waypoints was empty, or less than two before running, but no dice. I assume I have a misunderstanding in how all of this works, and I was not sure where to go further with debugging. So I come here, in assumption that this is not the intended behavior.

```
Error at java.base/java.lang.Thread.getStackTrace(Thread.java:1619): x and y components of Rotation2d are zero
        at java.base/java.lang.Thread.getStackTrace(Thread.java:1619)
        at edu.wpi.first.math.geometry.Rotation2d.<init>(Rotation2d.java:126)
        at edu.wpi.first.math.geometry.Rotation2d.rotateBy(Rotation2d.java:274)
        at edu.wpi.first.math.geometry.Rotation2d.plus(Rotation2d.java:213)
        at edu.wpi.first.math.geometry.Rotation2d.interpolate(Rotation2d.java:380)
        at com.pathplanner.lib.trajectory.PathPlannerTrajectory.cosineInterpolate(PathPlannerTrajectory.java:786)
        at com.pathplanner.lib.trajectory.PathPlannerTrajectory.generateStates(PathPlannerTrajectory.java:246)
        at com.pathplanner.lib.trajectory.PathPlannerTrajectory.<init>(PathPlannerTrajectory.java:69)
        at com.pathplanner.lib.path.PathPlannerPath.generateTrajectory(PathPlannerPath.java:1152)
[...]

PR #1061, opened 2025-02-12
Prevent crash when generating a trajectory from a one-state path

dspeyrer:
When a path is generated which only contains a single state (`PathPlannerPath.numPoints() == 1`), generating a trajectory crashes, as seen in #1056. This is caused by the interpolation between holonomic drivetrain rotations, which is trying to interpolate between two identical points, which have a distance of 0 apart from each other, thus causing a divide-by-zero.

When two path rotation targets fall on the same path point (as would be seen when there is only one point, and perhaps could be created by some other malformed path), instead of trying to interpolate between them, this fix just uses the rotation value for the next point.


mjansen4857:
I don't think this fully fixes the issue, there's probably other parts of the generation that expect more than one state. The best thing to do here is probably to have a special case at the beginning of the generation the checks for a single or zero state path and create a single trajectory state with the start pose/velocity etc, then skip the rest of the generation. This bypasses all possible issues with the normal trajectory generation.

mjansen4857:
Superseded by #1084

PR #1099, opened 2025-04-05
[Python] Fix rotation targets not being set when mirroring paths

PickleFace5:
[Unfortunately found this the hard way](https://www.youtube.com/watch?v=LjV7mkNZipc) :(

When mirroring paths, the rotation targets weren't being properly mirrored, which resulted in an AttributeError when generating the states for rotation targets that, well, don't have targets.


pathplannerlib-python/pathplannerlib/path.py
```diff
@@ -935,24 +935,28 @@ def mirrorPath(self) -> PathPlannerPath:
...
             PathPoint(self._mirrorTranslation(p.position))
             for p in self._allPoints
         ]
         for i, p in enumerate(self._allPoints):
-            path._allPoints[i].distanceAlongPath = p.distanceAlongPath
-            path._allPoints[i].maxV = p.maxV
-            path._allPoints[i].constraints = p.constraints
-            path._allPoints[i].waypointRelativePos = p.waypointRelativePos
-        if self._rotationTargets is not None:
-            path._rotationTargets = [RotationTarget(t.waypointRelativePosition, -t.target) for t in self._rotationTargets]
```
mjansen4857, line 949: I believe copying the rotation targets over is needed for something but can't remember what. Best to just leave that in

mjansen4857 (changes requested):

pathplannerlib-python/pathplannerlib/path.py
PickleFace5: See line 922, that change removed a duplicate assignment.

mjansen4857 (approved):

Issue #1100, opened 2025-04-10
PathPlannerPath::samplePath() can attempt to access a bad optional


Cynosure-null:
**Describe the bug**
Occasionally, when calling `PathPlannerPath::samplePath()` on a short path with only a start and end waypoint defined by the user, the robot code will crash. 

**To Reproduce**
See https://www.chiefdelphi.com/t/pathplanner-throwing-a-bad-optional-access-exception-on-path-construction/498617/2 for more info (example code, callstacks, etc.)

**Expected behavior**
The code won't crash.

**Versions: (please complete the following information):**
 - OS: Arch Linux
 - GUI Version: 2025.2.2
 - PPLib Version: 2025.2.2
 - PPLib Language: C++
**Additional context**
The issue occurs on line 929 of `pathplanner/pathplannerlib/src/main/native/cpp/pathplanner/lib/path/PathPlannerPath.cpp` due to calling `.value()` without first checking for validity. I am more than happy to create a PR for this issue, but I don't know what the desired behaviour should be if a cubic lerp cannot be created due to too few paths.

mjansen4857:
It should be impossible for those optionals to be accessed incorrectly as long as the data being fed in is valid. The issue here lies with whatever is creating the waypoints, which in your case is the waypointsFromPoses method. The only thing I see wrong with that currently is that you're passing in robot poses instead of creating a pose from position/heading. That shouldn't keep the waypoints from being valid but it would create a bad path.

mjansen4857:
I don't see any errors in the waypointsFromPoses logic. Are you manually creating waypoints anywhere?

Cynosure-null:
We are doing the following to initialize our path
```
                                   std::vector<Waypoint> waypoints = PathPlannerPath::waypointsFromPoses({m_odometry->getPose(),
                                                                                                           goal_pose});

                                    // The constraints for the path. TODO CHANGE IT
                                    PathConstraints constraints(1_mps, 2_mps_sq, 90_deg_per_s, 180_deg_per_s_sq);

                                    frc::ChassisSpeeds speeds = m_odometry->getFieldRelativeSpeeds();
                                    units::meters_per_second_t vel = static_cast<units::meters_per_second_t>(MathUtils::pythag(speeds.vx(), speeds.vy()));

                                    auto path = std::make_shared<PathPlannerPath>(
                                        waypoints,
                                        constraints,
[...]

PR #1101, opened 2025-04-10
improved error logging of PathPlannerPath::samplePath

aolgeirson:
#1100 added logging, however a non-crashing outcome would be ideal as the exception won't be handled

### DIFF
--- a/pathplannerlib/src/main/native/cpp/pathplanner/lib/path/PathPlannerPath.cpp
+++ b/pathplannerlib/src/main/native/cpp/pathplanner/lib/path/PathPlannerPath.cpp
@@ -924,10 +924,24 @@ frc::Translation2d PathPlannerPath::samplePath(
 
 	double t = pos - i;
 
-	auto p1 = m_waypoints[i].anchor;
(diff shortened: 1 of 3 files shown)


pathplannerlib/src/main/native/cpp/pathplanner/lib/path/PathPlannerPath.cpp
```diff
@@ -924,10 +924,24 @@ frc::Translation2d PathPlannerPath::samplePath(
...
-	auto p4 = m_waypoints[i + 1].anchor;
+	if (!m_waypoints.at(i).nextControl) {
+		throw std::runtime_error(
+				"PathPlannerPath::samplePath: Waypoint " + std::to_string(i)+ " .nextControl is null");
+	}
+
+	if (!m_waypoints.at(i+1).nextControl) {
+		throw std::runtime_error(
+				"PathPlannerPath::samplePath: Waypoint " + std::to_string(i+1)+ " .nextControl is null");
+	}
```
mjansen4857, line 935: This would always fail for every path. The end waypoint does not have a nextControl point

PR #1113, opened 2025-08-08
Add systemcore platform and remove roboRIO platform

spacey-sooty:


mjansen4857:
I believe the platform name is “linuxsystemcore” not just “systemcore”. 

Also, the platforms are correct in the hosted vendordep file, so these shouldn’t affect that issue. Those files haven’t even been updated to the 2027 version yet.

mjansen4857:
Unless they pulled the file directly from the example, in which case it needs to be swapped out entirely with the hosted version: https://3015rangerrobotics.github.io/pathplannerlib/PathplannerLibSystemCoreAlpha.json

spacey-sooty:
Should probably still be updated here but yeah your right I got the platform name wrong

PR #1115, opened 2025-08-30
Navgrid Based Start Tolerance for Pathfinding

c-eid:
Changed the minimum start distance from a hardcoded 0.5 to the diagonal of the current navgrid size.


mjansen4857:
Needs C++ and python. I can help if needed

c-eid:
> Finally, the check for being within 2 meters should remain 2 meters 

I just assumed that was necessary because I seen another PR with that in it

PR #1157, opened 2026-03-25
vendordeps: Update WPILibNewCommands

crueter:
Fails to link on Alpha 2 otherwise


JosephTLockwood:
Bumping this -- the C++ linker error here (`undefined reference to frc::EdgeCounterFilter`) is also blocking #1169 and presumably every other open PR. Any chance this can land?

PR #1167, opened 2026-04-22
GUI: Add curvature combs 

NukeMinecart:
Adds the ability to view the curvature of a path. Can be used to make smoother paths


NukeMinecart:
Some images of the combs
image (4)

PR #1171, opened 2026-05-17
feat: add AutoBuilder.configureDistanceBased for distance-based mode

JosephTLockwood:
## Summary
Third and final PR in the distance-based path-following stack. **Depends on #1169 and #1170** -- until those merge, this PR's diff includes their changes too.

Adds four overloads of \`AutoBuilder.configureDistanceBased(...)\` that mirror the existing time-based \`configure(...)\` parameter shape:

| Overload | output type | endConditions |
|---|---|---|
| 1 (full) | \`BiConsumer<ChassisSpeeds, DriveFeedforwards>\` | explicit |
| 2 | \`BiConsumer<...>\` | uses \`EndConditions.defaults()\` |
| 3 | \`Consumer<ChassisSpeeds>\` | explicit |
| 4 | \`Consumer<ChassisSpeeds>\` | uses \`EndConditions.defaults()\` |

Sets the same \`globals\` fields as \`configure(...)\` and builds a \`pathFollowingCommandBuilder\` that creates \`FollowPathDistanceCommand\`. Explicitly leaves \`pathfindingConfigured = false\` since \`PathfindingCommand\` is intrinsically time-based; integrating distance-based pathfinding is a separate concern.

## Why
First-class API parity with \`configure(...)\` so users can opt into distance-based mode with one call instead of wiring \`configureCustom\` themselves.

## Tests
None added. PathPlannerLib has zero existing \`AutoBuilder\` unit tests because \`RobotConfig\`'s static initializer transitively loads WPILib NetworkTables JNI, which isn't available in pure-Java test scope. Following the same convention; verified via compile + \`spotlessJavaCheck\`.

\`followPath(PathPlannerPath)\` and the rest of the public \`AutoBuilder\` surface work transparently with distance-based mode -- they all delegate to the \`pathFollowingCommandBuilder\` lambda which now produces \`FollowPathDistanceCommand\`. The \`pathfindToPose\` / \`pathfindThenFollowPath\` family correctly throws when used in distance-based mode (the standard not-configured error).

## Test plan
- [x] Existing 43 tests pass
- [x] \`spotlessJavaCheck\` clean
- [ ] End-to-end smoke test in a real robot project (in progress)

### DIFF
--- a/pathplannerlib/src/main/java/com/pathplanner/lib/auto/AutoBuilder.java
+++ b/pathplannerlib/src/main/java/com/pathplanner/lib/auto/AutoBuilder.java
@@ -140,6 +140,175 @@ public class AutoBuilder {
   }
 
   /**
+   * Configures the AutoBuilder to follow paths by projected arc length instead of elapsed time. The
+   * robot's pose is projected onto the path each cycle, and the controller targets the state at
+   * that arc length. Useful when disturbances (obstacles, contact, momentary slippage) would cause
+   * a time-based reference to race ahead of the actual robot progress.
+   *
+   * <p>Pathfinding is <strong>not</strong> configured by this method — distance-based mode supports
+   * direct path following only. For pathfinding workflows, use the time-based {@link
+   * #configure(Supplier, Consumer, Supplier, BiConsumer, PathFollowingController, RobotConfig,
+   * BooleanSupplier, Subsystem...)} method.
+   *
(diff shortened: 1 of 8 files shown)


JosephTLockwood:
@mjansen4857 Everything seems to be working now. Ended up coming up with a solution to the stop points. Should be good to review now.

